PR #19, opened 2015-09-07
Moves logging.properties to etc folder in Debian package install. Add…

paweldomas:
…s the possibility to log to SysLog using agafua-syslog lib.


simbo1905:
Closing. The proposed `-Werror` mechanism turned out to be unworkable and would not have achieved the goal:

1. **Project-wide `-Werror` fails the whole build** — the legacy codebase (`SipInfoJsonProtocol`, `JvbConference`, `Statistics`, `ActionServicesHandler`, etc.) emits hundreds of pre-existing `unchecked` warnings from json-simple raw types. `-Xlint:all -Werror` fails compilation on all of them, not just Scaleway code.
2. **javac has no lint category for unused private methods** — so `-Werror` would never have surfaced the dead `transcribe()` method regardless.

Instead the actual dead code was removed directly:
- unused private `transcribe(TranscriptionRequest, Consumer<TranscriptionResult>)` in `ScalewayVoxtralService`
- duplicate `import java.util.concurrent.Executors;`

`mvn clean test` green (42 tests pass).

PR #24, opened 2016-01-20
Fix missing colon in sed call_control replace.

reixd:
During postinstall jigasi replaces the call control value in the jitsi-meet config. But depending on the jitsi-meet config it may break the JSON format, due to the missing colon and the end o f the line.


damencho:
If there is already a comma at the end, with this change a second comma will be added? And by default there is comma after the commented call_control.

damencho:
Current logic is working with default configuration.

bgrozev:
> If there is already a comma at the end, with this change a second comma will be added?

I think it will, but [trailing commas are allowed in JS](http://stackoverflow.com/a/7246662), so it shouldn't cause any problems.

reixd:
You can use a dict like this
```javascript
d = { 
  a : foo,
  b : bar, 
};
```

It won't break anything and is allowed as stated by @bgrozev .
In some scenarios when using a customized jitsi-meet config.js, installing jigasi afterwards will break the config, due to the missing colon.

Issue #29, opened 2016-04-07
No audio on calls


damencho:
Creating an issue to track progress of fixing jigasi to work with latest jvb verison.
Since latest changes in videobridge jigasi stopped working. It used to work without DTLS, which seems a problem now. In order to investigate and fix it. jigasi needs first to be mavenized. A branch is pushed with on going work on that effort. In my todo list there are few jitsi bundles that need to be mavenized:
protocol-sip.jar

packetlogging.jar
httputil.jar
whack.jar
argdelegation-service.jar

hid-service.jar
hid.jar

After finishing that and having a build using maven we can continue with investigating the problem.

ibauersachs:
I would assume that this also depends on https://github.com/jitsi/jitsi/issues/193

damencho:
Thanks, yeah that is correct.

mcguinn:
Can you add this to list of no-audio as it relates to SIP-only conferences and SBCs that wont relay media until they see media from both side of the call.
https://github.com/jitsi/jigasi/issues/30

petervnv:
I also wanted to confirm I'm having the same no audio behaviour on latest jvb, jitmeet and jigasi debian 8 nightly packages.

arag00rn:
We are facing same issues here:

cat /etc/debian_version 8.4

ii  jicofo                         1.0-252-1                   amd64        JItsi Meet COnference FOcus
ii  jigasi                         1.0-99                      amd64        Jitsi Gateway for SIP
ii  jitsi-meet                     1.0.979-1                   all          WebRTC JavaScript video conferences
ii  jitsi-meet-prosody             1.0.979-1                   all          Prosody configuration for Jitsi Meet
ii  jitsi-videobridge              727-1                       amd64        WebRTC compatible Selective Forwarding Unit (SFU)

destroy-everything:
+1

mrgloba:
I have same issues:

cat /etc/debian_version 8.4

ii  jicofo 1.0-254-1 amd64
ii  jigasi 1.0-99 amd64
ii  jitsi-meet 1.0.985-1 all
ii  jitsi-meet-prosody 1.0.985-1 all
ii  jitsi-videobridge 729-1 amd64

In jigasi log
```
SEVERE: [602] net.sf.fmj.media.Log.error()   Unable to handle format: LINEAR, 48000.0 Hz, 16-bit, Mono, LittleEndian, Signed
[...]

PR #33, opened 2016-05-18
Mavenize and fix no-audio issue

damencho:


damencho:
Hey @emcho or @ibauersachs can you take a look and merge this one. 
All of the commits are for mavenizing jigasi, modifying pom.xml and just the last on is tweaking xmpp account properties which enables dtls and media started to flow which fixes issue #29 . I was testing the debian package creation is also working. Thanks!

PR #35, opened 2016-06-04
net.java.sip.communicator.service.gui.ALWAYS_TRUST_MODE_ENABLED is se…

psaavedra:
net.java.sip.communicator.service.gui.ALWAYS_TRUST_MODE_ENABLED is set in sip-communicator.properties

I could suggest this trivial change in the jigasi code as consequence of the topic discussed on this thread: http://lists.jitsi.org/pipermail/users/2016-June/011191.html (http://lists.jitsi.org/pipermail/users/2016-June/thread.html#11191)

The reason: move the configuration of  the ALWAYS_TRUST_MODE_ENABLED property to the sip-communicator.properties, so as sysadmin you can take the decision of trust or not in all the certificates. By default, I considered the ALWAYS_TRUST_MODE_ENABLED=false as a more secure option in order to prevent MiM attacks.

### DIFF
--- a/src/main/java/org/jitsi/jigasi/xmpp/CallControlComponent.java
+++ b/src/main/java/org/jitsi/jigasi/xmpp/CallControlComponent.java
@@ -122,6 +122,16 @@ public class CallControlComponent
         ConfigurationService config
             = ServiceUtils.getService(
                     bundleContext, ConfigurationService.class);
+        
+        Boolean always_trust_mode = config.getBoolean(
+                "net.java.sip.communicator.service.gui.ALWAYS_TRUST_MODE_ENABLED",false);
+        if (always_trust_mode)
+        {
+                // Always trust mode - prevent failures because there's no GUI
+                // to ask the user, but do we always want to trust so, in this
+                // mode, the service is vulnerable to Man-In-The-Middle attacks.
+                logger.warn("Always trust in remote TLS certificates mode is enabled");
+        }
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/jitsi/jigasi/xmpp/CallControlComponent.java
```diff
@@ -122,6 +122,16 @@ public void start(BundleContext bundleContext)
         ConfigurationService config
             = ServiceUtils.getService(
                     bundleContext, ConfigurationService.class);
+        
+        Boolean always_trust_mode = config.getBoolean(
+                "net.java.sip.communicator.service.gui.ALWAYS_TRUST_MODE_ENABLED",false);
```
damencho: This changes the current default behaviour.

damencho:

src/main/java/org/jitsi/jigasi/xmpp/CallControlComponent.java
damencho: This changes the current default behaviour. By default this should be enabled.

ibauersachs: @damencho I strongly disagree with that. The default should be secure. If someone needs to connect with TLS insecurely, he should have the option to set that.

damencho: Yep, I see your point and I agree.

psaavedra:
Signed.

PR #47, opened 2017-01-17
Adds an option to enable rtp relay to the sip side.

damencho:
By default it is disabled, enabling it will forward multiple ssrcs.


damencho:
Do not merge till we update jitsi-protocol and jitsi-protocol-media versions in the pom file.

damencho:
This approach is not working, closing PR.

PR #48, opened 2017-04-11
Muc call control

damencho:


src/main/java/org/jitsi/jigasi/xmpp/CallControlComponent.java
```diff
@@ -277,84 +215,19 @@ protected IQ handleIQSetImpl(IQ iq)
...
-                org.jivesoftware.smack.packet.IQ result
-                    = org.jivesoftware.smack.packet.IQ.createResultIQ(smackIq);
-
-                return IQUtils.convert(result);
-            }
-            else
-            {
-                return super.handleIQSet(iq);
-            }
+            logger.warn("Call controller not initialized");
```
paweldomas: it should return XMPP Error service unavailable

src/main/java/org/jitsi/jigasi/xmpp/CallControlMucActivator.java
```diff
@@ -0,0 +1,422 @@
...
+
+import org.jivesoftware.smack.*;
+import org.jivesoftware.smack.filter.*;
+import org.jivesoftware.smack.packet.*;
+import org.osgi.framework.*;
+
+import java.util.*;
+
+/**
+ * @author Damian Minkov
```
paweldomas: Some general description is missing. How this works when it's being activate etc.

src/main/java/org/jitsi/jigasi/xmpp/CallControlMucActivator.java
```diff
@@ -0,0 +1,422 @@
...
+        {
+            logger.warn("MUC call control disabled.");
+            return;
+        }
+
+        osgiContext.addServiceListener(this);
+
+        Collection<ServiceReference<ProtocolProviderService>> refs
+            = ServiceUtils.getServiceReferences(
+            osgiContext,
```
paweldomas: indentation seems to be wrong

src/main/java/org/jitsi/jigasi/xmpp/CallControlMucActivator.java
```diff
@@ -0,0 +1,422 @@
...
+     */
+    @Override
+    public void stop(BundleContext bundleContext)
+        throws Exception
+    {
+        osgiContext.removeServiceListener(this);
+
+        Collection<ServiceReference<ProtocolProviderService>> refs
+            = ServiceUtils.getServiceReferences(
+            osgiContext,
```
paweldomas: indentation

src/main/java/org/jitsi/jigasi/xmpp/CallControlComponent.java
damencho: This is handled by the super.... calls, I will test it once again, but I'm sure I tested this case and was seeing the errors in the client.

src/main/java/org/jitsi/jigasi/xmpp/CallControlMucActivator.java
```diff
@@ -0,0 +1,422 @@
...
+        }
+    }
+
+    /**
+     * Stops and unregister everything - providers, listeners.
+     * @param bundleContext
+     * @throws Exception
+     */
+    @Override
+    public void stop(BundleContext bundleContext)
```
paweldomas: this.callControl reference is not cleared. Also does it require to release any resources ?

src/main/java/org/jitsi/jigasi/xmpp/CallControlMucActivator.java
damencho: Ups, again. Will fix it, sorry.

src/main/java/org/jitsi/jigasi/xmpp/CallControlMucActivator.java
```diff
@@ -0,0 +1,422 @@
...
+        if (service instanceof ProtocolProviderService)
+        {
+            initializeNewProvider((ProtocolProviderService) service);
+        }
+        else if (service instanceof SipGateway)
+        {
+            SipGateway gateway = (SipGateway) service;
+            gateway.addSipGatewayListener(this);
+
+            if (this.callControl == null)
```
paweldomas: if this.callControl !== null it would be good to check for reference equality here and print na error to detect multiple SipGateway instances. I suppose we expect to have only one instance at a time ?

paweldomas: no, it doesn;t make sense what I wrote - deleting comment

src/main/java/org/jitsi/jigasi/xmpp/CallControlMucActivator.java
```diff
@@ -0,0 +1,422 @@
...
+            // sends initial stats, used some kind of advertising
+            // so jicofo can recognize us as real jigasi and load balance us
+            updatePresenceStatusForXmppProvider(pps);
+        }
+        catch (Exception e)
+        {
+            logger.error(e, e);
+        }
+
+        // getting direct access to the xmpp connection in order to add
```
paweldomas: why would you want to continue adding packet listener if the code above has failed ?

src/main/java/org/jitsi/jigasi/xmpp/CallControlMucActivator.java
```diff
@@ -0,0 +1,422 @@
...
+
+        int participants = 0;
+        for(GatewaySession ses : gateway.getActiveSessions())
+        {
+            participants += ses.getJvbChatRoom().getMembersCount();
+        }
+
+        Collection<ServiceReference<ProtocolProviderService>> refs
+            = ServiceUtils.getServiceReferences(
+            osgiContext,
```
paweldomas: indentation

src/main/java/org/jitsi/jigasi/xmpp/CallControlMucActivator.java
```diff
@@ -0,0 +1,422 @@
...
+        if (ProtocolNames.JABBER.equals(pps.getProtocolName())
+            && pps.getAccountID() instanceof JabberAccountID
+            && !((JabberAccountID)pps.getAccountID()).isAnonymousAuthUsed())
+        {
+            try
+            {
+                String roomName = pps.getAccountID()
+                    .getAccountPropertyString(ROOM_NAME_ACCOUNT_PROP);
+                if (roomName == null)
+                {
```
paweldomas: { not on new line

src/main/java/org/jitsi/jigasi/xmpp/CallControlMucActivator.java
```diff
@@ -0,0 +1,422 @@
...
+    @Override
+    public void onJvbRoomJoined(GatewaySession source)
+    {
+        updatePresenceStatusForXmppProviders();
+    }
+
+    @Override
+    public void onSessionAdded(GatewaySession session)
+    {
+        session.setListener(this);
```
paweldomas: there's no need to call updatePresenceStatusForXmppProvider() here ?

src/main/java/org/jitsi/jigasi/xmpp/CallControlComponent.java
damencho: You were right, it needs return null; to signal 501 feature not implemented.

paweldomas: yes but the error will be feature not implemented or something like that

paweldomas: it should be "service unavailable" not "feature not implemented"

src/main/java/org/jitsi/jigasi/xmpp/CallControlMucActivator.java
damencho: Nope, no resources. CallControl is just a utility shared between CallControlComponent and CallControlMucActivator. Nothing to clear in both places.

src/main/java/org/jitsi/jigasi/xmpp/CallControlMucActivator.java
damencho: Yep, cause we haven't joined the jvb room, so we can take the participants count. Will add a comment there.

src/main/java/org/jitsi/jigasi/xmpp/CallControlMucActivator.java
damencho: ?

paweldomas: lol, forget about it I've mixed up JS with Java

src/main/java/org/jitsi/jigasi/xmpp/CallControlComponent.java
```diff
@@ -219,6 +219,11 @@ public IQ handleIQ(IQ iq)
         if (callControl == null)
         {
             logger.warn("Call controller not initialized");
+
+            // send service unavailable
+            IQ error = IQ.createResultIQ(iq);
```
paweldomas: will it be IQ type of error if createResultIQ is used ?

damencho: Yep, when you do iq.setError it sets the type of iq to error and then adds the error (I saw the code from tinder library itself). And what I receive in the client is:

<iq xmlns="jabber:client" type="error" to="" from="successfulbullscommunicateeverywhere@conference.damencho.jitsi.net/focus" id=""><error code="503" type="CANCEL"><service-unavailable xmlns="urn:ietf:params:xml:ns:xmpp-stanzas"/></error></iq>

src/main/java/org/jitsi/jigasi/xmpp/CallControlMucActivator.java
damencho: You are not alone. I've been moving { on new lines at some point in all of the code :)

PR #49, opened 2017-04-27
Custom bosh url

damencho:


src/main/java/org/jitsi/jigasi/GatewaySession.java
```diff
@@ -44,6 +45,19 @@
     private final static Logger logger = Logger.getLogger(GatewaySession.class);
 
     /**
+     * Property name of extra INVITE header which specifies password required
+     * to enter MUC room that is hosting the Jitsi Meet conference.
+     */
+    public String JITSI_MEET_ROOM_PASS_HEADER = "Jitsi-Conference-Room-Pass";
```
paweldomas: Why is field named like constant ? It seems to also serve the role of a default value, but then it should be a static field DEFAULT_JITSI_MEET_ROOM_PASS_HEADER.

Please separate into field and static default value.

src/main/java/org/jitsi/jigasi/GatewaySession.java
```diff
@@ -58,6 +58,21 @@
         = "JITSI_MEET_ROOM_PASS_HEADER_NAME";
 
     /**
+     * Property name of optional INVITE header which specifies the base domain
+     * to be used to extract the subdomain from the roomname in order
+     * to construct custom bosh URL to enter MUC room that is hosting
+     * the Jitsi Meet conference.
+     */
+    public String JITSI_MEET_DOMAIN_BASE_HEADER
```
paweldomas: It seems strange to use a field as constant and a default value at the same time.

src/main/java/org/jitsi/jigasi/GatewaySession.java
```diff
@@ -168,6 +184,12 @@ public GatewaySession(SipGateway gateway)
             .getAccountPropertyString(
                 JITSI_MEET_ROOM_PASS_HEADER_PROPERTY,
                 JITSI_MEET_ROOM_PASS_HEADER);
+
+        // check for custom header name for domain base header
+        JITSI_MEET_DOMAIN_BASE_HEADER = sipProvider.getAccountID()
```
paweldomas: This is hard to understand, because it's a field and a default value at the same time.

It's an interesting idea, but I don't think we use this anywhere else in our Java code.

PR #55, opened 2017-07-23
add initial transcription module, set language level to 1.8, add GoogleCloudTranscriptionService

nikvaessen:
This PR relies on[ RawStreamListener from libjitsi](https://github.com/nikvaessen/libjitsi/tree/RawStreamListener)

### DIFF
--- a/src/main/java/org/jitsi/jigasi/transcription/GoogleCloudTranscriptionService.java
+++ b/src/main/java/org/jitsi/jigasi/transcription/GoogleCloudTranscriptionService.java
@@ -1,0 +1,526 @@
+/*
+ * Jigasi, the JItsi GAteway to SIP.
+ *
+ * Copyright @ 2017 Atlassian Pty Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jitsi.jigasi.transcription;
+
+import com.google.api.gax.grpc.ApiStreamObserver;
+import com.google.api.gax.grpc.StreamingCallable;
+import com.google.cloud.speech.spi.v1.SpeechClient;
+import com.google.cloud.speech.v1.*;
+import com.google.protobuf.ByteString;
+import org.jitsi.util.Logger;
+
+import javax.media.format.AudioFormat;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.function.Consumer;
+
+/**
+ * Implements a TranscriptionService which will use Google Cloud
+ * speech-to-text API to transcribe audio
+ *
+ * Requires google cloud sdk on the machine running the JVM running
+ * this application. See
+ * https://cloud.google.com/speech/docs/getting-started
+ * for details
+ *
+ * @author Nik Vaessen
+ */
+public class GoogleCloudTranscriptionService
+    implements TranscriptionService
+{
+
+    /**
+     * BCP-47 (see https://www.rfc-editor.org/rfc/bcp/bcp47.txt)
+     * language tags of the languages supported by Google cloud speech
+     * to-text API (See https://cloud.google.com/speech/docs/languages)
+     */
+    public final static String[] SUPPORTED_LANGUAGE_TAGS = new String[]
+        {
+            "af-ZA",
+            "id-ID",
+            "ms-MY",
+            "ca-ES",
+            "cs-CZ",
+            "da-DK",
+            "de-DE",
+            "en-AU",
+            "en-CA",
+            "en-GB",
+            "en-IN",
+            "en-IE",
+            "en-NZ",
+            "en-PH",
+            "en-ZA",
+            "en-US",
+            "es-AR",
+            "es-BO",
+            "es-CL",
+            "es-CO",
+            "es-CR",
+            "es-EC",
+            "es-SV",
+            "es-ES",
+            "es-US",
+            "es-GT",
+            "es-HN",
+            "es-MX",
+            "es-NI",
+            "es-PA",
+            "es-PY",
+            "es-PE",
+            "es-PR",
+            "es-DO",
+            "es-UY",
+            "es-VE",
+            "eu-ES",
+            "fil-PH",
+            "fr-CA",
+            "fr-FR",
+            "gl-ES",
+            "hr-HR",
+            "zu-ZA",
+            "is-IS",
+            "it-IT",
+            "lt-LT",
+            "hu-HU",
+            "nl-NL",
+            "nb-NO",
+            "pl-PL",
+            "pt-BR",
+            "pt-PT",
+            "ro-RO",
+            "sk-SK",
+            "sl-SI",
+            "fi-FI",
+            "sv-SE",
+            "vi-VN",
+            "tr-TR",
+            "el-GR",
+            "bg-BG",
+            "ru-RU",
+            "sr-RS",
+            "uk-UA",
+            "he-IL",
+            "ar-IL",
+            "ar-JO",
+            "ar-AE",
+            "ar-BH",
+            "ar-DZ",
+            "ar-SA",
+            "ar-IQ",
+            "ar-KW",
+            "ar-MA",
+            "ar-TN",
+            "ar-OM",
+            "ar-PS",
+            "ar-QA",
+            "ar-LB",
+            "ar-EG",
+            "fa-IR",
+            "hi-IN",
+            "th-TH",
+            "ko-KR",
+            "cmn-Hant-TW",
+            "yue-Hant-HK",
+            "ja-JP",
+            "cmn-Hans-HK",
+            "cmn-Hans-CN",
+        };
+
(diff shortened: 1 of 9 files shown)


bgrozev:
I would like to make some changes to the names in the API that we added to libjitsi:
RawStreamListener -> ReceiveStreamBufferListener
samplesRead -> handleBuffer
Other suggestions would be welcome, I just really dislike RawStreamListener and samplesRead (which I came up with...) because they are so specific. 

Could you open a PR in libjitsi with those changes?

src/main/java/org/jitsi/jigasi/transcription/GoogleCloudTranscriptionService.java
```diff
@@ -0,0 +1,655 @@
...
+import com.google.cloud.speech.v1.*;
+import com.google.protobuf.ByteString;
+import org.jitsi.util.Logger;
+
+import javax.media.format.AudioFormat;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.function.Consumer;
```
bgrozev: Star imports please

src/main/java/org/jitsi/jigasi/transcription/GoogleCloudTranscriptionService.java
```diff
@@ -0,0 +1,655 @@
...
+     */
+    private final static Logger logger
+            = Logger.getLogger(GoogleCloudTranscriptionService.class);
+
+    /**
+     * ExecutorService which will manage the async requests of audio
+     * transcriptions. CashedThreadPool is chosen because sending
+     * audio chunks are small short-lived async tasks.
+     */
+    private ExecutorService executorService = Executors.newCachedThreadPool();
```
bgrozev: Do we need to worry about the requests being sent in order?

src/main/java/org/jitsi/jigasi/transcription/TranscriptionService.java
```diff
@@ -0,0 +1,120 @@
...
+     * transcription to be successful. Fragments longer than 60 seconds
+     * might not be supported by most speech-to-text or will require
+     * to much processing time to get a timely answer.
+     *
+     * @param request the TranscriptionRequest which holds the audio
+     * @param resultConsumer a Consumer of the transcription result
+     * @throws UnsupportedOperationException when this service does not support
+     * fragmented audio speech-to-text
+     */
+    void sent(TranscriptionRequest request,
```
bgrozev: Can we rename this to e.g. sendRequest?

nikvaessen: I will rename it `sendSingleRequest` to emphasise that this method should be used for one audio complete file/fragment

src/main/java/org/jitsi/jigasi/transcription/GoogleCloudTranscriptionService.java
nikvaessen: No, not for `sendSingleRequest()`

src/main/java/org/jitsi/jigasi/transcription/GoogleCloudTranscriptionService.java
nikvaessen: Should I always use start imports, or only when I have x of the same package?

src/main/java/org/jitsi/jigasi/transcription/Transcriber.java
```diff
@@ -0,0 +1,550 @@
...
+
+        executorService.submit(() ->
+        {
+            // ReceiveStream uses unsigned longs
+            long ssrc = receiveStream.getSSRC() & 0xffffffffL;
+
+            Participant p = participants.get(ssrc);
+            if(p != null)
+            {
+                participants.get(ssrc).giveBuffer(buffer);
```
bgrozev: Why call get() a second time when you already have the result in "p"?

src/main/java/org/jitsi/jigasi/transcription/Transcriber.java
```diff
@@ -0,0 +1,550 @@
...
+        /**
+         * The size of the local buffer. A single packet is expected to contain
+         * 1920 bytes, so the size should be a multiple of 1920
+         */
+        private final int BUFFER_SIZE = EXPECTED_AUDIO_LENGTH * 25;
+
+        /**
+         * Whether we should buffer locally before sending
+         */
+        private final Boolean USE_LOCAL_BUFFER = true;
```
bgrozev: Perhaps make these static? A little explanation of where the numbers come from would be useful: with webrtc we usually see 20ms opus frames which are decoded to 2 bytes per sample and 48000Hz sampling rate.

src/main/java/org/jitsi/jigasi/transcription/Transcriber.java
```diff
@@ -0,0 +1,550 @@
...
+         */
+        private String name;
+
+        /**
+         * The id identifying audio belonging to this participant
+         */
+        private long ssrc;
+
+        /**
+         * The streaming
```
bgrozev: This looks incomplete

src/main/java/org/jitsi/jigasi/transcription/Transcriber.java
```diff
@@ -0,0 +1,550 @@
...
+            }
+        }
+
+        /**
+         * When a participant left it does not accept audio and thus no new
+         * results will come in
+         */
+        void left()
+        {
+            session.end();
```
bgrozev: I would add a null check just in case

src/main/java/org/jitsi/jigasi/transcription/Transcriber.java
bgrozev: Oh, and the aim is to have the buffer store 500ms worth of audio, right?

bgrozev:
Merged the PR in libjitsi. You can update the API and use the new maven version: libjitsi-1.0-20170724.205625-298

src/main/java/org/jitsi/jigasi/transcription/GoogleCloudTranscriptionService.java
```diff
@@ -0,0 +1,642 @@
...
+
+    /**
+     * Sends audio as an array of bytes to speech-to-text API of google cloud
+     *
+     * @param request the TranscriptionRequest which holds the audio to be sent
+     * @param resultConsumer a Consumer which will handle the
+     *                       TranscriptionResult
+     */
+    @Override
+    public void sentSingleRequest(final TranscriptionRequest request,
```
bgrozev: Do you mind renaming to "sendSingleRequest" (with a "d")?

src/main/java/org/jitsi/jigasi/transcription/Transcriber.java
bgrozev: Here we give a reference to "buffer" to another thread, and then we return from samplesRead() which runs in a mixer thread. So two threads have "buffer", which could be a problem (if e.g. the mixer thread decides to re-use the buffer before "our" thread has a chance to run and make a copy. I think the code which makes the copy should run in the mixer thread, and it should submit the task only when it needs to send send a request (this will also have less tasks to submit, since not every Buffer results in a request being sent).

src/main/java/org/jitsi/jigasi/transcription/GoogleCloudTranscriptionService.java
```diff
@@ -0,0 +1,642 @@
...
+            catch(Exception e)
+            {
+                e.printStackTrace();
+            }
+        }
+
+        @Override
+        public void give(final TranscriptionRequest request)
+        {
+            this.service.submit(() -> requestManager.sentRequest(request));
```
bgrozev: Can we rename to sendRequest (with a "d")?

src/main/java/org/jitsi/jigasi/transcription/Transcriber.java
nikvaessen: > Perhaps make these static?
Made them static and moved the Locale up because inner classes can't have static declarations. 

> A little explanation of where the numbers come from would be useful: with webrtc we usually see 20ms opus frames which are decoded to 2 bytes per sample and 48000Hz sampling rate.
added

> Oh, and the aim is to have the buffer store 500ms worth of audio, right?
The aim is indeed to store 500ms to save bandwith. Although this might've been the reason I was seeing some errors yesterday ("The audio was not being received real-time"). Maybe it should be a bit lower

src/main/java/org/jitsi/jigasi/transcription/Transcriber.java
nikvaessen: fixed

src/main/java/org/jitsi/jigasi/transcription/Transcriber.java
nikvaessen: fixed

src/main/java/org/jitsi/jigasi/transcription/GoogleCloudTranscriptionService.java
nikvaessen: fixed

src/main/java/org/jitsi/jigasi/transcription/Transcriber.java
nikvaessen: > Why call get() a second time when you already have the result in "p"?
fixed

> I think the code which makes the copy should run in the mixer thread
 and it should submit the task only when it needs to send send a request (this will also have less tasks to submit, since not every Buffer results in a request being sent)

Okay, I've moved the local buffering to the mixer thread.

src/main/java/org/jitsi/jigasi/transcription/GoogleCloudTranscriptionService.java
nikvaessen: I'm not sure sendRequest is a right name either, because we do not necessarily immediately send a request, but I've changed it

PR #56, opened 2017-07-23
Add a TranscriptionGateway and TranscriptionGatewaySession

nikvaessen:
relies on [PR 55](https://github.com/jitsi/jigasi/pull/55) in jigasi 
relies on changes in [jitsi-protocol and jitsi-protocol-jabber](https://github.com/jitsi/jitsi/tree/transcription)

### DIFF
--- a/src/main/java/org/jitsi/jigasi/JvbConference.java
+++ b/src/main/java/org/jitsi/jigasi/JvbConference.java
@@ -268,35 +268,6 @@ public class JvbConference
         sendPresenceExtension(mediaPresence);
     }
 
-    private String getMucDisplayName()
-    {
-        String mucDisplayName = null;
-
-        if(gatewaySession instanceof SipGatewaySession)
-        {
-            String sipDestination = callContext.getDestination();
-            Call sipCall = ((SipGatewaySession) gatewaySession).getSipCall();
-
-            if (sipDestination != null)
-            {
-                mucDisplayName = sipDestination;
-            }
-            else if (sipCall != null)
-            {
-                CallPeer firstPeer = sipCall.getCallPeers().next();
-                if (firstPeer != null)
-                {
-                    mucDisplayName = firstPeer.getDisplayName();
-                }
-            }
-        }
-        if(gatewaySession instanceof TranscriptionGatewaySession)
-        {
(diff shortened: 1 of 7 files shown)


bgrozev:
The jitsi changes are in, you can use these versions:
jitsi-protocol-2.11-20170724.170155-4
jitsi-protocol-jabber-2.11-20170724.170120-10

src/main/java/org/jitsi/jigasi/JvbConference.java
```diff
@@ -291,6 +290,10 @@ else if (sipCall != null)
                 }
             }
         }
+        if(gatewaySession instanceof TranscriptionGatewaySession)
+        {
+            mucDisplayName = TranscriptionGatewaySession.DISPLAY_NAME;
+        }
```
bgrozev: Thanks for renaming this! I would refactor a bit more and move the method to GatewaySession (it already has the callContext) to avoid casting.

src/main/java/org/jitsi/jigasi/TranscriptionGateway.java
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * A Gateway which creates a TranscriptionGatewaySession when it has an outgoing call
+ *
+ * @author Nik Vaessen
+ */
+public class TranscriptionGateway
+    extends AbstractGateway<TranscriptionGatewaySession>
+{
+
+    public TranscriptionGateway(BundleContext context)
```
bgrozev: Please add javadocs. Our convention requires them on every method.

bgrozev: Even if it is just "{@inheritDoc}"

nikvaessen: Just to make sure, using "@Override" will assume "{@inheritdoc}" when there is no comment?

src/main/java/org/jitsi/jigasi/JvbConference.java
nikvaessen: fixed

src/main/java/org/jitsi/jigasi/TranscriptionGateway.java
bgrozev: OK

PR #57, opened 2017-07-23
added a TranscriptionGateway OSGI service

nikvaessen:
relies on [PR 56](https://github.com/jitsi/jigasi/pull/56)

When the DialIQ's uri is "transcribe:", a transcriber will join the room, otherwise a sip connection joins the room

### DIFF
--- a/src/main/java/org/jitsi/jigasi/JigasiBundleActivator.java
+++ b/src/main/java/org/jitsi/jigasi/JigasiBundleActivator.java
@@ -17,8 +17,6 @@
  */
 package org.jitsi.jigasi;
 
-import java.util.*;
-
 import net.java.sip.communicator.impl.protocol.jabber.*;
 import net.java.sip.communicator.impl.protocol.jabber.extensions.jitsimeet.*;
 import net.java.sip.communicator.impl.protocol.jabber.extensions.rayo.*;
@@ -29,6 +27,8 @@ import net.java.sip.communicator.util.*;
 import org.jitsi.service.configuration.*;
 import org.jivesoftware.smack.provider.*;
 import org.osgi.framework.*;
+
+import java.util.*;
 
 /**
  * Jigasi bundle activator. Registers {@link SipGateway} and waits for the first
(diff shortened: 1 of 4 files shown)


src/main/java/org/jitsi/jigasi/JigasiBundleActivator.java
```diff
@@ -50,7 +53,8 @@
 
     public static BundleContext osgiContext;
 
-    private SipGateway gateway;
+    private SipGateway sipGateway;
+    private TranscriptionGateway transcriptionGateway;
```
bgrozev: Can you please add javadocs and add yourself as author?

src/main/java/org/jitsi/jigasi/JigasiBundleActivator.java
```diff
@@ -104,7 +111,7 @@ public void stop(BundleContext bundleContext)
     {
         logger.info("Stopping JigasiBundleActivator");
 
-        gateway.stop();
+        sipGateway.stop();
```
bgrozev: I know that TranscriptionGateway#stop() doesn't do anything right now, but I think we should call it here anyway.

src/main/java/org/jitsi/jigasi/xmpp/CallControl.java
```diff
@@ -146,23 +191,33 @@ else if (allowedJid == null)
 
                 String callResource = ctx.getCallResource();
 
-                gateway.createOutgoingCall(ctx);
+                if("transcribe:".equals(to))
```
bgrozev: Can we move this to a constant somewhere, and name it something that is even less likely to use used as a sip address (e.g. jitsi_transcribe)? This will be hidden by a UI somewhere anyway.

src/main/java/org/jitsi/jigasi/JigasiBundleActivator.java
nikvaessen: I feel a bit hesitant to add javadocs to stuff I did not write, I've added to the gateways but I don't know how to document osgiContect and UiserviceStub

src/main/java/org/jitsi/jigasi/JigasiBundleActivator.java
nikvaessen: fixed

src/main/java/org/jitsi/jigasi/JigasiBundleActivator.java
bgrozev: Yeah, that's fine.

PR #62, opened 2017-07-27
Alternatives

bgrozev:
Do not merge yet please

### DIFF
--- a/src/main/java/org/jitsi/jigasi/transcription/TranscriptionResult.java
+++ b/src/main/java/org/jitsi/jigasi/transcription/TranscriptionResult.java
@@ -17,59 +17,122 @@
  */
 package org.jitsi.jigasi.transcription;
 
+import java.util.*;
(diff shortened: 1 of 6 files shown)


src/main/java/org/jitsi/jigasi/transcription/TranscriptionResult.java
```diff
@@ -17,85 +17,167 @@
...
+     *
+     * @return the name
+     */
+    public String getName()
+    {
+        return participant == null
+            ? Participant.UNKNOWN_NAME
+            : participant.getName();
+    }
+
```
nikvaessen: When rebasing, there should be a method getId(). This should now use the participant. Also, 
in getName(), should UNKNOWN_NAME be changed to the id?

nikvaessen:
Changes seem good to me, waiting for rebase for a final check

PR #63, opened 2017-07-28
Store final transcripts locally as .txt files

nikvaessen:


src/main/java/org/jitsi/jigasi/transcription/Transcriber.java
```diff
@@ -127,30 +126,52 @@
...
+    public Transcriber(String roomName, TranscriptionService service)
     {
         if(!service.supportsStreamRecognition())
         {
             throw new IllegalArgumentException("Currently only services which" +
-                    " support streaming recognition are supported");
+                " support streaming recognition are supported");
         }
         this.transcriptionService = service;
+        addTranscriptionListener(this.transcript);
```
bgrozev: Should we also handle roomName?

src/main/java/org/jitsi/jigasi/transcription/Transcriber.java
```diff
@@ -192,8 +217,13 @@ public void start()
...
         if(State.NOT_STARTED.equals(this.state))
         {
+            logger.debug("Transcriber is now transcribing");
             this.state = State.TRANSCRIBING;
-            executorService = Executors.newSingleThreadExecutor();
+            this.executorService = Executors.newSingleThreadExecutor();
+
+            List<String> names = new ArrayList<>(participants.size());
+            participants.values().forEach((p) -> names.add(p.getName()));
+            this.transcript.started(roomName, names);
```
bgrozev: When we list the participants, the ones without a name should be listed with IDs. Perhaps we can handle this by modifying Participant#getName, or handle it here?

src/main/java/org/jitsi/jigasi/transcription/Transcriber.java
```diff
@@ -336,6 +371,30 @@ public AudioMixerMediaDevice getMediaDevice()
     }
 
     /**
+     * If all participant have been completely transcribed, and state is
+     * FINISHING_UP, the state can be set to FINISHED
+     */
+    private void isFinishedUp()
```
bgrozev: This is confusing -- the name starts with "is", but it doesn't return a boolean? Can you please rename and clarify the doc?

src/main/java/org/jitsi/jigasi/transcription/Transcript.java
```diff
@@ -0,0 +1,427 @@
...
+    implements TranscriptionListener
+{
+    /**
+     * Formats an Instant to <hour:minute:second> in the timezone the machine
+     * is running on
+     */
+    private static final DateTimeFormatter timeFormatter
+        = DateTimeFormatter
+        .ofLocalizedTime(FormatStyle.MEDIUM)
+        .withZone(ZoneId.systemDefault());
```
bgrozev: I think we should either use UTC, or explicitly include the timezone somewhere. What do you think?

src/main/java/org/jitsi/jigasi/transcription/Transcript.java
```diff
@@ -0,0 +1,427 @@
...
+
+    /**
+     * Notify the transcript that the conference started at this exact moment
+     * Will be ignored if already told the conference started
+     *
+     * @param initialNames the names of the participants currently in the room
+     */
+    public void started(List<String> initialNames)
+    {
+        if(started != null)
```
bgrozev: Shouldn't this be started==null?

src/main/java/org/jitsi/jigasi/transcription/LocalTxtTranscriptHandler.java
```diff
@@ -0,0 +1,413 @@
...
+    /**
+     * The logger of this class
+     */
+    private static final Logger logger = Logger.getLogger(Transcript.class);
+
+    /**
+     * The root dir where a new directory will be created in to store
+     * transcripts
+     */
+    private static final String ROOT_DIR = System.getProperty("user.dir");
```
bgrozev: Can you use the config service and a separate property name, e.g. org.jitsi.jigasi.transcription.LocalTxtTranscriptHandler.DIRECTORY (which can combine ROOT_DIR and LOG_DIR)?

src/main/java/org/jitsi/jigasi/transcription/LocalTxtTranscriptHandler.java
```diff
@@ -0,0 +1,413 @@
...
+                return;
+            }
+        }
+
+        File t = new File(logDir, "transcript_" + Instant.now() + ".txt");
+
+        try(FileWriter writer = new FileWriter(t))
+        {
+            writer.write(transcript);
+            System.out.println(transcript);
```
bgrozev: Can we just print one line that acknowledges that we saved it in a file?

src/main/java/org/jitsi/jigasi/transcription/Transcriber.java
nikvaessen: fixed

src/main/java/org/jitsi/jigasi/transcription/Transcriber.java
nikvaessen: fixed

src/main/java/org/jitsi/jigasi/transcription/Transcriber.java
nikvaessen: fixed

src/main/java/org/jitsi/jigasi/transcription/Transcript.java
nikvaessen: fixed

src/main/java/org/jitsi/jigasi/transcription/LocalTxtTranscriptHandler.java
nikvaessen: org.jitsi.jigasi.transcription.LocalTxtTranscriptHandler.DIRECTORY can now be added to sip-communicator.properties

src/main/java/org/jitsi/jigasi/transcription/LocalTxtTranscriptHandler.java
nikvaessen: fixed

src/main/java/org/jitsi/jigasi/transcription/Transcript.java
nikvaessen: we decided on UTC in meetings if I recall correctly, so set it to that

PR #64, opened 2017-08-01
more gracefully fail when Google Cloud credentials are not set correctly

nikvaessen:
warn in log and sends a message to the chat before leaving immediately
wait for merge until I tested it on a computer with the credentials set

### DIFF
--- a/src/main/java/org/jitsi/jigasi/TranscriptionGatewaySession.java
+++ b/src/main/java/org/jitsi/jigasi/TranscriptionGatewaySession.java
@@ -122,6 +122,16 @@ public class TranscriptionGatewaySession
         this.jvbCall = jvbConferenceCall;
         this.chatRoom = super.jvbConference.getJvbRoom();
 
+        // If the transcription service is not correctly configured, there is no
+        // point in continuing this session, so end it immediately
+        if(!service.isConfiguredProperly())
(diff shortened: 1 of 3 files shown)


src/main/java/org/jitsi/jigasi/TranscriptionGatewaySession.java
```diff
@@ -122,6 +122,16 @@ Exception onConferenceCallStarted(Call jvbConferenceCall)
...
         this.chatRoom = super.jvbConference.getJvbRoom();
 
+        // If the transcription service is not correctly configured, there is no
+        // point in continuing this session, so end it immediately
+        if(!service.isConfiguredProperly())
+        {
+            logger.warn("TranscriptionService is not properly set");
+            sendMessageToRoom("Transcriber is not properly configured. " +
+                    "Contact the host and let him now! I will now leave");
+            jvbConference.stop();
```
bgrozev: Shouldn't we return here, before we add members and start the transcriber?

How about changing the chat message to something like "Contact the service administrators and let them know"?

nikvaessen: fixed, and tested on a machine with credentials and everything was okay

PR #65, opened 2017-08-02
Refactor GoogleCloudTranscriptionService to better deal with 1 minute StreamingRecognize restriction

nikvaessen:
Includes: 
add constants for maximum alternatives interim results and single utterance
remove overlapping sessions
create new session whenever a final result comes in
close session whenever there is an error or a final result


nikvaessen:
UPDATE:
added a thread which cancels a session if a participant muted their audio

PR #67, opened 2017-08-03
Alternatives2

bgrozev:


bgrozev:
@nikvaessen Please take a look. I print the confidence intervals by default, perhaps we can make that configurable in the future.

src/main/java/org/jitsi/jigasi/transcription/Transcriber.java
```diff
@@ -137,14 +130,14 @@
...
+     * the audio streams
      */
     public Transcriber(String roomName, TranscriptionService service)
     {
-        if(!service.supportsStreamRecognition())
+        if (!service.supportsStreamRecognition())
         {
             throw new IllegalArgumentException("Currently only services which" +
-                " support streaming recognition are supported");
+                                                   " support streaming recognition are supported");
```
nikvaessen: does it exceed the 80 chars limit here?

src/main/java/org/jitsi/jigasi/transcription/Transcriber.java
```diff
@@ -200,43 +193,44 @@ public void add(String name, String id, long ssrc)
...
-                if((name = p.getName()) == null)
-                {
-                    name = p.getId();
-                }
-                names.add(name);
-            });
+            participants.values().forEach((p) ->
+                                          {
+                                              String name;
+                                              if ((name = p.getName()) == null)
```
nikvaessen: We don't need to do this anymore as p.getName() should never return null

src/main/java/org/jitsi/jigasi/transcription/TranscriptionAlternative.java
```diff
@@ -0,0 +1,79 @@
...
+ * An alternative transcription of a chunk of audio.
+ *
+ * @author Boris Grozev
+ */
+public class TranscriptionAlternative
+{
+    /**
+     * The default confidence value.
+     */
+    private static double DEFAULT_CONFIDENCE = 1.0D;
```
nikvaessen: Wouldn't it be better if we only print confidence if it's not 1?

src/main/java/org/jitsi/jigasi/transcription/TranscriptionResult.java
```diff
@@ -17,85 +17,169 @@
...
+    }
+
+    /**
+     * Get the name of the participant saying the audio
+     *
+     * @return the name
+     */
+    public String getName()
+    {
+        return participant == null
```
nikvaessen: Not necessary anymore as participant.getName() should never return null

src/main/java/org/jitsi/jigasi/transcription/TranscriptionResult.java
```diff
@@ -17,85 +17,169 @@
...
+     * {@inheritDoc}
+     * </p>
+     * In this default implementation we include all alternative transcriptions.
+     */
     @Override
     public String toString()
     {
-        return name == null ?
-                transcription : name + ": " + transcription;
+        String name = getName();
```
nikvaessen: Same as above

src/main/java/org/jitsi/jigasi/transcription/TranscriptionResult.java
```diff
@@ -17,85 +17,169 @@
...
+        {
+            name = Participant.UNKNOWN_NAME;
+        }
+
+        StringBuilder sb = new StringBuilder(name);
+        sb.append(": ");
+        if (!alternatives.isEmpty())
+        {
+            sb.append('[')
+                .append(alternatives.get(0).getConfidence())
```
nikvaessen: See note about default confidence

nikvaessen:
You added a lot of white space between if, for statements and the opening bracket "(", is this a code convention?

src/main/java/org/jitsi/jigasi/transcription/TranscriptionResult.java
bgrozev: But this check is for TranscriptionResult#participant

src/main/java/org/jitsi/jigasi/transcription/Transcriber.java
bgrozev: Fixed

src/main/java/org/jitsi/jigasi/transcription/Transcriber.java
bgrozev: Fixed (which also removed the weird indentation, it was introduced by my IDE and I failed to notice)

src/main/java/org/jitsi/jigasi/transcription/TranscriptionAlternative.java
bgrozev: This constant was meant as a default when the confidence or a result is not known (e.g. if another s2t API is used)

src/main/java/org/jitsi/jigasi/transcription/TranscriptionResult.java
bgrozev: Fixed

bgrozev:
> You added a lot of white space between if, for statements and the opening bracket "(", is this a code convention?

In this case that was the automatic formatting that my IDE did and I didn't notice. But, yes, we have a convention for using a space after "if" and "for".

bgrozev:
About printing the confidence: at this stage, I only want it printed so that we can look at it during testing and get a better idea of how trustworthy it is (e.g. should we just replace results with confidence <X with "<could not be transcribed>"?). So for this use-case, I don't think not printing the confidence when it's "1" will be any use. 

But I'll add a commit that makes printing of the confidence configurable with a default of false.

bgrozev:
Actually, I just remembered why I didn't put a property yesterday -- because this will be sent to the client in a structured way, so it will very soon be obsolete. I'll keep it as it is right now.

src/main/java/org/jitsi/jigasi/transcription/TranscriptionResult.java
nikvaessen: You're right, sorry

nikvaessen:
Tested the branch and everything seems to work after the fix :)

PR #74, opened 2017-08-14
Refactor how transcript are saved to more easily serve them with nginx/jetty

nikvaessen:


src/main/java/org/jitsi/jigasi/TranscriptionGateway.java
```diff
@@ -21,14 +21,44 @@
...
+     * Whether to send results in json to
+     * {@link net.java.sip.communicator.service.protocol.ChatRoom} of muc
+     */
+    private final static boolean SEND_RESULT_JSON = false;
+
+    /**
+     * Whether to send results in txt to
+     * {@link net.java.sip.communicator.service.protocol.ChatRoom} of muc
+     */
+    private final static boolean SEND_RESULT_TXT = true;
```
bgrozev: Why are these final? Shouldn't they be configurable with properties?

src/main/java/org/jitsi/jigasi/transcription/AbstractTranscriptPublisher.java
```diff
@@ -0,0 +1,477 @@
...
+
+    /**
+     * Get a file name for a transcript, which includes the time and some ID
+     * to make it hard to guess
+     *
+     * @return the file name
+     */
+    protected String getHardToGuessFileName()
+    {
+        return "transcript_" + UUID.randomUUID() + "_" + Instant.now();
```
bgrozev: Since this is an instance method of a specific publisher, wouldn't it make sense to save the generated value and return it in subsequent calls to getHardToGuessFileName()? Otherwise, I would make it static and rename to "generate" instead of "get" for clarity.

src/main/java/org/jitsi/jigasi/TranscriptionGateway.java
nikvaessen: made them configurable

src/main/java/org/jitsi/jigasi/transcription/AbstractTranscriptPublisher.java
nikvaessen: changed the name and made it static

bgrozev: And just a convention thing: static methods should go up in the file (before the instance fields).

nikvaessen: fixed

src/main/java/org/jitsi/jigasi/transcription/TranscriptHandler.java
```diff
@@ -17,130 +17,103 @@
...
     /**
-     * The Formatter which used a fluent API
+     * Get the {@link TranscriptPublisher.Promise}s which can handle a
+     * {@link Transcript} for all given {@link TranscriptPublisher}s
      *
-     * @param <T> the type wherein the transcript will be formatted
+     * @return a list of {@link TranscriptPublisher.Promise}s
      */
-    interface Formatter<T>
+    public List<TranscriptPublisher.Promise> getTranscriptPublishPromises()
```
bgrozev: This was a little confusing since I didn't realize that each publisher creates a new promise.

src/main/java/org/jitsi/jigasi/TranscriptionGateway.java
```diff
@@ -54,10 +129,54 @@ public TranscriptionGatewaySession createOutgoingCall(CallContext ctx)
...
+    /**
+     * Get whether to save transcripts in txt
+     *
+     * @return true if saved in txt, false otherwise
+     */
+    private boolean getStoreInTxt()
+    {
+        return JigasiBundleActivator.getConfigurationService()
+            .getBoolean(P_NAME_SAVE_TRANSCRIPT_TXT, SAVE_TRANSCRIPT_TXT);
+    }
```
bgrozev: I like that the configuration is isolated to one file, but I wonder if TranscriptionGateway is the best place for it. Do you think it would make sense to move it to TranscriptHandler?

src/main/java/org/jitsi/jigasi/TranscriptionGateway.java
```diff
@@ -21,13 +21,67 @@
...
 public class TranscriptionGateway
     extends AbstractGateway<TranscriptionGatewaySession>
 {
+
+    /**
+     * Property name for saving transcript in json
+     */
+    public final static String P_NAME_SAVE_TRANSCRIPT_JSON
+        = "org.jitsi.jigasi.TranscriptionGateway.SAVE_TRANSCRIPT_JSON";
+    /**
```
bgrozev: How about changing this to "org.jitsi.jigasi.transcription.SAVE_JSON"?

bgrozev:
The overall design looks good to me -- it's easy to add a new implementation and enable it via configuration properties. I've left a couple more comments inline.

src/main/java/org/jitsi/jigasi/TranscriptionGateway.java
nikvaessen: So we would make the adding the interfaces private to the TranscriptHandler?

src/main/java/org/jitsi/jigasi/transcription/TranscriptHandler.java
nikvaessen: I don't really see a great way how to do it otherwise... How would you combine multiple descriptions?

src/main/java/org/jitsi/jigasi/TranscriptionGateway.java
nikvaessen: and the other one SAVE_TXT?

bgrozev: Yes, change the prefix on all of them

src/main/java/org/jitsi/jigasi/TranscriptionGateway.java
bgrozev: Yeah, we would just have TranscriptHandler configure itself when instantiated. This is just a suggestion, though, feel free to use whatever more makes sense to you.

src/main/java/org/jitsi/jigasi/transcription/TranscriptHandler.java
bgrozev: Sorry, I was unclear. I meant that the documentation could be improved (and perhaps the naming, although I don't really have a suggestion).

nikvaessen: I updated the documentation a bit. I hope it's more clear now

src/main/java/org/jitsi/jigasi/TranscriptionGateway.java
nikvaessen: I moved it to the TranscriptHandler as it seems a bit cleaner

src/main/java/org/jitsi/jigasi/TranscriptionGateway.java
nikvaessen: fixed

src/main/java/org/jitsi/jigasi/transcription/TranscriptHandler.java
```diff
@@ -31,6 +32,52 @@
...
+     * Property name for sending result in json
+     */
+    public final static String P_NAME_SEND_JSON
+        = "org.jitsi.jigasi.TranscriptionGateway.SEND_JSON";
+
+    /**
+     * Property name for sending result in txt
+     */
+    public final static String P_NAME_SEND_TXT
+        = "org.jitsi.jigasi.TranscriptionGateway.SEND_TXT";
```
bgrozev: Can we change the prefixes, please?

nikvaessen: Yes, we can!

nikvaessen:
added a push which plays together with the changes in lib-jitsi-meet 
https://github.com/jitsi/lib-jitsi-meet/pull/575#issuecomment-323327055

PR #76, opened 2017-08-18
Add a jetty bundle to serve transcripts

nikvaessen:


src/main/java/org/jitsi/jigasi/rest/TranscriptServerBundleActivator.java
```diff
@@ -0,0 +1,92 @@
...
+     * this {@link AbstractJettyBundleActivator}.
+     */
+    public static final String JETTY_PROPERTY_PREFIX
+        = "org.jitsi.jigasi.transcription";
+
+    /**
+     * Property name for the port which will be used by this Jetty instance.
+     */
+    public static final String P_NAME_TRANSCRIPT_PORT
+        = "org.jitsi.jigasi.transcription.PORT";
```
bgrozev: AbstractJettyBundleActivator already provides configuration of the port via properties (with JETTY_PROPERTY_PREFIX + ".jetty.port").

src/main/java/org/jitsi/jigasi/rest/TranscriptServerBundleActivator.java
```diff
@@ -0,0 +1,92 @@
...
+        super(JETTY_PROPERTY_PREFIX);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    protected int getDefaultPort()
+    {
+        return JigasiBundleActivator.getConfigurationService().getInt(
```
bgrozev: Just return -1 here (and in getDefaultTlsPort), AbstractJettyBundleActivator will pick up the value through configuration if it is set:
https://github.com/jitsi/jicoco/blob/master/src/main/java/org/jitsi/rest/AbstractJettyBundleActivator.java#L335

PR #82, opened 2017-08-29
Update to Smack 4.2

ibauersachs:


src/main/java/org/jitsi/jigasi/xmpp/CallControl.java
```diff
@@ -146,99 +159,91 @@ public CallControl(ConfigurationService config)
...
+     * @return an {@link IQ} stanza which represents the response to the
+     * specified request.
+     */
+    public IQ handleHangUp(HangUp iq)
+        throws CallControlAuthorizationException
+    {
+        checkAuthorized(iq);
 
-                SipGatewaySession session = sipGateway.getSession(callResource);
+        // FIXME: 23/07/17 Transcription sessions should also be able
```
bbaldino: anything here that needs to be addressed before merging?

ibauersachs: This must be a mismatched merge, the note isn't from me.

ibauersachs: Actually not a mismatched merge. The note was there before, just at a different position (in the if/multi-else type check of handleIQ).

nikvaessen:
This PR should be ready to be reviewed again.

damencho:
Me and @nikvaessen we've been testing for a while with this branch, so if no one objects I will merge it tomorrow morning (CDT).

PR #85, opened 2017-09-12
Add transcription status to presence

hristoterezov:


src/main/java/org/jitsi/jigasi/xmpp/TranscriptionStatusExtension.java
```diff
@@ -0,0 +1,100 @@
...
+ * <tt>{@link Status#OFF}</tt> - transcription stopped(available to start)
+ * <tt>{@link Status#ON}</tt> - transcription in progress
+ */
+public class TranscriptionStatusExtension
+        extends AbstractPacketExtension
+{
+    /**
+     * The namespace of this packet extension.
+     */
+    public static final String NAMESPACE = "http://jitsi.org/protocol/transcription";
```
bgrozev: NOOOOOO :)

src/main/java/org/jitsi/jigasi/TranscriptionGatewaySession.java
```diff
@@ -550,6 +551,14 @@ public String getMucDisplayName()
         return TranscriptionGatewaySession.DISPLAY_NAME;
     }
 
+    @Override
+    public void notify(Transcriber transcriber, TranscriptEvent event) {
+        TranscriptionStatusExtension extension = new TranscriptionStatusExtension();
+        extension.setStatus( transcriber.isTranscribing()
+                ? TranscriptionStatusExtension.Status.ON : TranscriptionStatusExtension.Status.OFF);
+        jvbConference.sendPresenceExtension(extension);
```
bgrozev: It looks like this will update its presence with every new transcription result. Or am I missing something? I would expect it to update its presence when the status changes (when it starts and when it stops).

bgrozev: Actually, it implements TranscriptionEventListener but isn't registered anywhere as a listener?

damencho:
The resulting commits of this PR are now in master:
0ff46d494abb0bbf0f5803741991afcca8765ab9
7921f736bf39c1c08c667338f87182792de630c1

PR #86, opened 2017-09-27
Adds callstats integration for calls using jitsi-stats.

damencho:


src/main/java/org/jitsi/jigasi/stats/StatsHandler.java
```diff
@@ -0,0 +1,377 @@
...
+     * CallStats account property: jigasiId
+     * The jigasi id to report to callstats.io.
+     */
+    private static final String CS_ACC_PROP_JIGASI_ID
+        = "CallStats.jigasiId";
+
+    /**
+     * The default jigasi id to use if setting is missing.
+     */
+    public static final String DEFAULT_JIGASI_ID = "jigasi";
```
bgrozev: I wonder if we can move most of these (as well as their counterparts in jitsi-videobridge) to a common abstract class in jitsi-stats.

PR #105, opened 2018-05-29
refactor the way Participants are created and linked to the audio SSRC

nikvaessen:


src/main/java/org/jitsi/jigasi/AbstractGatewaySession.java
```diff
@@ -262,7 +262,7 @@ void notifyMemberJoined(ChatRoomMember member)
      * @param member the member who left the JVB conference
      */
     // FIXME: 17/07/17 JvbConference does not yet call this method
-    void notifyMemberLeft(ChatRoomMember member)
+    void notifyChatRoomMemberLeft(ChatRoomMember member)
```
bgrozev: Unrelated to this PR, but: why do we not update `this.participantsCount` here? It seems to only increase

nikvaessen: I don't know, it's supposed to be global:

```
    /**
     * Global participant count during this session including the focus.
     */
    private int participantsCount = 0;
```

src/main/java/org/jitsi/jigasi/transcription/Transcriber.java
```diff
@@ -170,45 +175,99 @@ public Transcriber(TranscriptionService service)
...
+            participant.joined();
+            TranscriptEvent event = transcript.notifyJoined(participant);
+            if (event != null)
+            {
+                fireTranscribeEvent(event);
+            }
+
+            if (logger.isDebugEnabled())
+                logger.debug(
+                    "Removed participant with identifier " + identifier);
```
bgrozev: removed?

src/main/java/org/jitsi/jigasi/transcription/Transcriber.java
```diff
@@ -170,45 +175,99 @@ public Transcriber(TranscriptionService service)
...
-     * @param chatMember the chat participant to be added
-     * @param ssrc the ssrc of the participant to be added
+     * @param identifier the identifier of the participant
      */
-    public void add(ChatRoomMember chatMember, long ssrc)
+    public void participantJoined(String identifier)
     {
-        Participant participant;
-        if (this.participants.containsKey(ssrc))
+        if (this.participants.containsKey(identifier))
```
bgrozev: This code will throw a NPE if the participant is removed between the `containsKey` check and the call to `get`. I would just call `get` once and then check for `null`.

src/main/java/org/jitsi/jigasi/TranscriptionGatewaySession.java
```diff
@@ -228,100 +231,51 @@ void onJvbConferenceWillStop(JvbConference jvbConference, int reasonCode,
...
-            logger.debug("Starting WaitThread for joining ChatRoomMember "+
-                    chatMember.getDisplayName());
-            new WaitForConferenceMemberThread(chatMember,
-                    this::addParticipant).start();
-        }
-        else
-        {
-            addParticipant(chatMember, confMember);
-        }
+        this.transcriber.maybeAddParticipant(identifier);
```
bgrozev: Why do we need 3 function calls here? Use one function and let the `Transcriber` figure it out

src/main/java/org/jitsi/jigasi/TranscriptionGatewaySession.java
```diff
@@ -370,118 +324,67 @@ public void toneReceived(DTMFReceivedEvent dtmfReceivedEvent)
      */
     private void addInitialMembers()
     {
-        List<ConferenceMember> confMembers = getCurrentConferenceMembers();
         List<ChatRoomMember> chatRoomMembers = getCurrentChatRoomMembers();
+        List<ConferenceMember> confMembers = getCurrentConferenceMembers();
 
-        if(confMembers == null || chatRoomMembers == null)
+        if(chatRoomMembers == null || confMembers == null)
```
bgrozev: Why do we want both to be not null? The code below handles them independently

bgrozev:
The access to `Transcriber.participants` is not thread safe. Do you know that it is only accessed by a single thread? My suspicion is that it's not

src/main/java/org/jitsi/jigasi/AbstractGatewaySession.java
bgrozev: I see

src/main/java/org/jitsi/jigasi/AbstractGatewaySession.java
```diff
@@ -261,9 +261,9 @@ void notifyChatRoomMemberJoined(ChatRoomMember member)
      *
      * @param member the member who left the JVB conference
      */
-    // FIXME: 17/07/17 JvbConference does not yet call this method
     void notifyChatRoomMemberLeft(ChatRoomMember member)
     {
+        participantsCount--;
```
bgrozev: Wait, didn't you say that it shouldn't decrement, because it counts the number of participants that joined the call, not the number of current participants?

nikvaessen: I cleared it up with damyan and it should indeed NOT decrement

damencho: Its me that brought the confusion thinking of some stats and without reading the javadoc where it is written:
```
     * Returns the cumulative number of participants that were active during
     * this session including the focus.
```

PR #106, opened 2018-06-10
add avatar-url and optionally <identity> information to transcription

nikvaessen:
Shouldn't be merged until version `2.13-SNAPSHOT` can be replaced with an actual release. Relies on PR https://github.com/jitsi/jitsi/pull/483

The JSON library is replaced due google's json-simple library always escaped `\`, making it impossible to put URL into JSON format. 

Currently pulling to smack4.2 branch, once that is merged this PR can point to master.


pom.xml
```diff
@@ -90,12 +90,12 @@
     <dependency>
       <groupId>${project.groupId}</groupId>
       <artifactId>jitsi-protocol</artifactId>
-      <version>${jitsi-desktop.version}</version>
+      <version>2.13-SNAPSHOT</version>
```
bgrozev: We can now use the new version, right?

src/main/java/org/jitsi/jigasi/CallContext.java
```diff
@@ -381,4 +381,25 @@ public Object getSource()
...
+    {
+        String url = getBoshURL();
+        String room = getRoomName();
+
+        if(url == null || room == null)
+        {
+            return null;
+        }
+
+        url = url.substring(0, url.indexOf("http-bind?"));
```
bgrozev: Ugh. This looks flaky. I think the "?" might not always be present. Can we do "/http-bind" and then add "/"? Can we also add a log message (it would make it easier to debug if something here brakes in the future).

src/main/java/org/jitsi/jigasi/xmpp/ChatRoomMemberPresenceExtensionReader.java
```diff
@@ -0,0 +1,167 @@
...
+
+import net.java.sip.communicator.impl.protocol.jabber.*;
+import net.java.sip.communicator.impl.protocol.jabber.extensions.jitsimeet.*;
+import net.java.sip.communicator.service.protocol.*;
+import org.jitsi.jigasi.util.*;
+import org.jitsi.util.*;
+import org.jivesoftware.smack.packet.*;
+
+/**
+ * This class is able to receive {@link Presence} objects and update data
```
bgrozev: We read the identity fields from the `ChatRoomMemberJabberImpl` presence and store them in the `ChatRoomMemberJabberImpl` in another format (using the `DataObject`) API.

Why don't we just define helper functions to read this data from presence?

nikvaessen: the lastPresence might be overwritten, so we need to either store all presences retrieved over the lifetime of the object, implement member fields in `ChatRoomMemberJabberImpl` with getters/setters, or use the DataObject API

src/main/java/org/jitsi/jigasi/transcription/Participant.java
```diff
@@ -190,7 +207,90 @@ public String getAvatarUrl()
...
+        String avatarId;
+        if((avatarId
+            = (String) memberJabber.getData(
+            ChatRoomMemberPresenceExtensionReader.AVATAR_ID)) != null)
+        {
+            return String.format(MEEPLE_URL_FORMAT,
+                Util.stringToMD5hash(avatarId));
+        }
+
+        return null;
```
bgrozev: Perhaps we can still return a valid random avatar based on the user id if everything else is missing?

src/main/java/org/jitsi/jigasi/transcription/Participant.java
```diff
@@ -190,7 +207,90 @@ public String getAvatarUrl()
...
+            ChatRoomMemberPresenceExtensionReader.IDENTITY_USERID
+        );
+    }
+
+    /**
+     * Get the group id in the identity presence, if present
+     *
+     * @return the group id or null
+     */
+    public String getIdentityGroupId()
```
bgrozev: Related to the other comment I left: these could read from the last received presence directly.

PR #108, opened 2018-06-13
Adds recording indications for on/off to sip participants.

damencho:
Works only with opus files and in translator mode.


src/main/java/org/jitsi/jigasi/AbstractGatewaySession.java
```diff
@@ -236,6 +238,17 @@ void notifyJvbRoomJoined()
...
     }
 
+    /**
+     *  Method called by {@link JvbConference} to notify session that it has
+     *  detected a recording status change.
+     *
+     * @param mode the recording mode.
+     * @param status the recording status.
+     */
+    void notifyForRecordingStatusChange(
```
bgrozev: I was going to suggest that we use `onRecordingStatusChanged` for consistency, but we have 4 methods named with "on" and 5 methods named with "notify"...
How about `notifyRecordingStatusChanged`?

src/main/java/org/jitsi/jigasi/SipGatewaySession.java
```diff
@@ -792,6 +840,15 @@ public void run()
                     }
                 }).start();
             }
+
+            // when someone connects and recording is on, play notification
+            if (CallPeerState.CONNECTED.equals(callPeerState))
+            {
+                if (currentRecordingStatus.equals(JibriIq.Status.ON))
+                {
+                    Util.injectSoundFile(call, "sounds/RecordingOn.opus");
```
bgrozev: Use the constant

src/main/java/org/jitsi/jigasi/util/Util.java
```diff
@@ -79,4 +83,101 @@ public static String extractCallIdFromResource(String callResource)
     {
         return callResource.substring(0, callResource.indexOf("@"));
     }
+
+    /**
+     * Injects as sound file in a call's <tt>MediaStream</tt> using injectPacket
```
bgrozev: Can you please clean up the doc?

src/main/java/org/jitsi/jigasi/SipGatewaySession.java
```diff
@@ -660,6 +676,38 @@ else if (sipCall != null)
...
+        if (currentRecordingStatus.equals(status))
+        {
+            return;
+        }
+
+        if (JibriIq.Status.ON.equals(status))
+        {
+            // if call is still not established this will be ignored in
+            // injectSoundFile and nothing will be played
+            Util.injectSoundFile(this.call, REC_ON_SOUND);
```
bgrozev: This is synchronous and takes about as long as the length of the audio file (i.e. seconds). Are we really OK blocking whatever thread fired the event for that long?

PR #113, opened 2018-06-21
fix npe check for avatar-id and rename method for formatting xml string

nikvaessen:


nikvaessen:
should update the jitsi-desktop-version to include the small changes (dataobject, constructor)

src/main/java/org/jitsi/jigasi/util/Util.java
```diff
@@ -262,7 +262,7 @@ public static String stringToMD5hash(String toHash)
      * @param indent the indent to apply to the XML
      * @return the formatted String
      */
-    public static String prettyFormat(String input, int indent)
+    public static String prettyFormatXMLString(String input, int indent)
```
bgrozev: Why do we have an unused method here?

nikvaessen: I've used it for debugging the presence objects stored in the ChatRoomMember. But is now unused as it was only used for a logging message in the deleted `ChatRoomMemberPresenceReader`

PR #114, opened 2018-06-29
WIP : Translates final transcript messages into hardcoded language

pvgupta24:
screenshot from 2018-06-29 13-56-16

- Uses the language code in `sip-communicator.properties` 
`org.jitsi.jigasi.transcription.TARGET_LANGUAGE=hi` field. 
- Translation is disabled if left empty.

### DIFF
--- a/src/main/java/org/jitsi/jigasi/transcription/GoogleCloudTranslationService.java
+++ b/src/main/java/org/jitsi/jigasi/transcription/GoogleCloudTranslationService.java
@@ -1,0 +1,65 @@
+/*
+ * Jigasi, the JItsi GAteway to SIP.
+ *
+ * Copyright @ 2018 Atlassian Pty Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jitsi.jigasi.transcription;
+
+import com.google.cloud.translate.*;
+import com.google.cloud.translate.Translate.*;
+
+/**
+ * Implements a {@link TranslationService} which will use Google Cloud
+ * translate API to translate the given text from one language to another.
+ *
+ * Requires google cloud sdk on the machine running the JVM running
+ * this application. See
+ * https://cloud.google.com/translate/docs/quickstart
+ * for details
+ *
+ * @author Praveen Kumar Gupta
+ */
+public class GoogleCloudTranslationService
+        implements TranslationService
+{
+
+    /**
+    * The translation model to be used for translation.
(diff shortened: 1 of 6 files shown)


src/main/java/org/jitsi/jigasi/transcription/GoogleCloudTranslationService.java
```diff
@@ -0,0 +1,65 @@
...
+    /**
+    * The translation model to be used for translation.
+    * Possible values are {@code base} and {@code nmt}
+    */
+    private final TranslateOption model = TranslateOption.model("nmt");
+
+    /**
+    * Translation service object for accessing the cloud translate API.
+    */
+    private static final Translate translator
```
bgrozev: I am a little confused by the fact that some of the fields are static while some are not. If you want to have a single `translator` instance, and also use the `TranslationService` interface, I would suggest using the singleton design pattern:
1. Initialize a single GoogleCloudTranslationService instance and save it in a static field (`private static T INSTANCE = new T()`).
2. Expose the instance through a static method (`public static T getInstance()`)
3. Make the constructor private

src/main/java/org/jitsi/jigasi/transcription/LocalTxtTranscriptHandler.java
```diff
@@ -55,6 +56,18 @@
...
+    /**
+     * Property name for getting a target language for translation.
+     */
+    public final static String P_NAME_TARGET_LANGUAGE
+            = "org.jitsi.jigasi.transcription.TARGET_LANGUAGE";
+
+    /**
+     * Whether to translate text before sending results and the target language.
+     */
+    public final static String TARGET_LANGUAGE
```
bgrozev: What do you think about renaming to "DEFAULT_TARGET_LANGUAGE"? I think it's more descriptive

src/main/java/org/jitsi/jigasi/transcription/LocalTxtTranscriptHandler.java
```diff
@@ -190,7 +205,11 @@ public void publish(ChatRoom chatRoom, TranscriptionResult result)
         String name = result.getName();
         String transcription = result.getAlternatives().iterator()
             .next().getTranscription();
-
+        if(!getTargetLanguage().isEmpty())
+        {
+            transcription
+                    = translationService.translate(transcription, "en","hi");
```
bgrozev: Any reason to hard-code "hi" here instead of using the value of `getTargetLanguage`?

src/main/java/org/jitsi/jigasi/transcription/LocalTxtTranscriptHandler.java
bgrozev: Or you could remove it and just use ""

src/main/java/org/jitsi/jigasi/transcription/TranslationManager.java
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jitsi.jigasi.transcription;
+
+public class TranslationManager {
```
bgrozev: Is this class used anywhere?

pvgupta24:
@bgrozev this PR was just had some initial results. I am making the updates in the new PR.

PR #116, opened 2018-07-02
Leave whack version to come from jicoco dependency.

damencho:
Seems the problem is using an older(2.0.0) whack version, while jicoco uses 2.0.1. This should fix it, although I don't see to be fixed locally. I keep seeing that it depends: org.jitsi:jicoco:jar:1.1-SNAPSHOT:compile which depends on whack 2.0.0. But anyway whack should be coming with jicoco and should not be declared in jigasi.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -50,6 +50,7 @@
     <dependency>
       <groupId>org.igniterealtime.whack</groupId>
       <artifactId>core</artifactId>
+      <version>2.0.1</version>
     </dependency>
     <dependency>
       <groupId>org.gagravarr</groupId>


damencho:
Hum, this 2.0.0 probably is coming from jitsi-universe-public :( need to think it, maybe remove it from there ...

saghul:
Not sure I'm qualified to review this :-O Maybe @nikvaessen can help?

damencho:
Seems the problem is using an older(2.0.0) whack version, while jicoco uses 2.0.1. This should fix it, specifying the exact version, but I think we should remove the whack version from jigasi, jvb and jicofo and this should be coming from jicoco. But we need also to update jitsi-universe-public and jitsi-universe.

paweldomas:
Yes, I guess that this version should be declared only in jicoco. There's no code in Jigasi which deals with whack directly.

damencho:
Yeah and this is the case and for jvb and jicofo :) Let's at least fix the build and will get there. Need to check all those maven parent projects and publish them.

nikvaessen:
LGTM

PR #119, opened 2018-07-03
make enabling gateways configurable

nikvaessen:


src/main/java/org/jitsi/jigasi/JigasiBundleActivator.java
```diff
@@ -54,6 +54,30 @@
 
     public static BundleContext osgiContext;
 
+    /**
+     * The property name for the boolean value whether transcription should be
+     * enabled.
+     */
+    public final static String P_NAME_ENABLE_TRANSCRIPTION
+        = "org.jitsi.jigasi.enableTranscription";
```
bgrozev: Can we rename to ENABLE_TRANSCRIPTION (and ENABLE_SIP) for consistency with the other properties that we have?

src/main/java/org/jitsi/jigasi/xmpp/CallControlMucActivator.java
```diff
@@ -347,15 +347,34 @@ public void onSessionFailed(AbstractGatewaySession session)
...
+            osgiContext, TranscriptionGateway.class);
 
         int participants = 0;
-        for(SipGatewaySession ses : gateway.getActiveSessions())
+        int sessions  = 0;
+
+        if(sipGateway != null)
         {
-            participants += ses.getJvbChatRoom().getMembersCount();
+            for(SipGatewaySession ses : sipGateway.getActiveSessions())
```
bgrozev: How about collecting all active sessions and then looping over them once (avoid duplicating the counting code)

src/main/java/org/jitsi/jigasi/xmpp/CallControlMucActivator.java
```diff
@@ -428,16 +447,34 @@ private void updatePresenceStatusForXmppProvider(
     private void updatePresenceStatusForXmppProvider(
```
bgrozev: Ugh. Why do we have `updatePresenceStatusForXmppProvider(pps)` and `updatePresenceStatusForXmppProviders()` that implement the same logic twice? :/

I know that's not related to the PR, so I don't expect you to fix it

nikvaessen:
```java
java.lang.NullPointerException
	at net.java.sip.communicator.impl.protocol.jabber.OperationSetBasicTelephonyJabberImpl$JingleIqSetRequestHandler.handleIQRequest(OperationSetBasicTelephonyJabberImpl.java:1009)
	at org.jivesoftware.smack.AbstractXMPPConnection$3.run(AbstractXMPPConnection.java:1153)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
```
Currently this exception occurs when sip is disabled and transcriber leaves room.

PR #121, opened 2018-07-04
only let transcriber use hidden domain when joining a conference

nikvaessen:
This PR should make the transcriber be able to join a conference with hidden domain while a SIP call would join as a normal participant.

### DIFF
--- a/src/main/java/org/jitsi/jigasi/AbstractGatewaySession.java
+++ b/src/main/java/org/jitsi/jigasi/AbstractGatewaySession.java
@@ -321,6 +321,18 @@ public abstract class AbstractGatewaySession
     public abstract boolean isTranslatorSupported();
 
     /**
+     * Whether the gateway should attempt to use an authenticated xmmp account
+     * to join the conference. If the properties for authenticated login are
+     * set, and this method returns true, authenticated login will happen.
+     * If the properties are not set, authenticated login will fail regardless.
+     * If this method returns false, authenticated login will not happen
+     * regardless of whether the properties are set.
+     *
+     * @return whether to attempt authenticated login
+     */
+    public abstract boolean maybeUseAuthenticatedLogin();
+
+    /**
      * Returns the gateway used for this session.
      * @return the gateway used for this session.
      */
--- a/src/main/java/org/jitsi/jigasi/AbstractGatewaySession.java
+++ b/src/main/java/org/jitsi/jigasi/AbstractGatewaySession.java
@@ -21,6 +21,7 @@ import net.java.sip.communicator.impl.protocol.jabber.extensions.jibri.*;
 import net.java.sip.communicator.service.protocol.*;
 import net.java.sip.communicator.service.protocol.event.*;
 import net.java.sip.communicator.util.*;
+import org.jivesoftware.smack.packet.*;
 
 import java.util.*;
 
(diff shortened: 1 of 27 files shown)


src/main/java/org/jitsi/jigasi/SipGatewaySession.java
```diff
@@ -618,6 +618,15 @@ public void toneReceived(DTMFReceivedEvent dtmfReceivedEvent)
         }
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public boolean maybeUseAuthenticatedLogin()
+    {
+        return false;
```
damencho: So this disables authenticated login for jigasi at all? Is this correct?

nikvaessen: Disabling it for dials using SIP

damencho: We should not do that, people are already using it.
What you can do is add an account property that disables auth on xmpp account and act base on that. This way it will be per sip account in case of many sip accounts that serve multiple and different deployments.

src/main/java/org/jitsi/jigasi/JvbConference.java
```diff
@@ -1094,67 +1094,73 @@ else if(jvbCall.getCallState() == CallState.CALL_ENDED)
...
-            else if ("org.jitsi.jigasi.xmpp.acc.BOSH_URL_PATTERN"
-                        .equals(overridenProp))
-            {
-                // do not override boshURL with the global setting if
-                // we already have a value
-                if (StringUtils.isNullOrEmpty(ctx.getBoshURL()))
-                    ctx.setBoshURL(value);
-            }
-            else
+        if(gatewaySession.maybeUseAuthenticatedLogin())
```
damencho: This can be `if (!useAuth) return;` to avoid formatting with one more tab/one more nested level.

src/main/java/org/jitsi/jigasi/AbstractGatewaySession.java
```diff
@@ -320,6 +320,18 @@ public int getParticipantsCount()
      */
     public abstract boolean isTranslatorSupported();
 
+    /**
+     * Whether the gateway should attempt to use an authenticated xmmp account
+     * to join the conference. If the properties for authenticated login are
+     * set, and this method returns true, authenticated login will happen.
+     * If the properties are not set, authenticated login will fail regardless.
```
bgrozev: What does "will fail" mean? Will not be attempted?

src/main/java/org/jitsi/jigasi/JvbConference.java
```diff
@@ -1094,6 +1094,18 @@ else if(jvbCall.getCallState() == CallState.CALL_ENDED)
...
+        AbstractGateway gw = gatewaySession.getGateway();
+        if(gw instanceof SipGateway)
+        {
+            String preventAuthLogin
+                = ((SipGateway) gw).getSipAccountProperty("PREVENT_AUTH_LOGIN");
+
+            if (Boolean.valueOf(preventAuthLogin))
+            {
+                return properties;
+            }
```
damencho: Returning the properties here will prevent using any custom overrides coming from org.jitsi.jigasi.xmpp.acc properties, not only the authentication.

PR #124, opened 2018-07-12
Sends json-messages for translations & Sets the participant translation language on presence update

pvgupta24:
Depends on 
- [x] [jitsi/jitsi#504](https://github.com/jitsi/jitsi/pull/504) for presence extension. `pom.xml` to be updated with the new `jitsi-dektop-version`

Can be tested by dialing the transcriber and setting the language from console : 
`APP.conference._room.setLocalParticipantProperty('translation_language','hi');`

### DIFF
--- a/src/main/java/org/jitsi/jigasi/TranscriptionGatewaySession.java
+++ b/src/main/java/org/jitsi/jigasi/TranscriptionGatewaySession.java
@@ -41,7 +41,8 @@ import java.util.*;
 public class TranscriptionGatewaySession
     extends AbstractGatewaySession
     implements TranscriptionListener,
-               TranscriptionEventListener
+               TranscriptionEventListener,
+               TranslationResultListener
 {
     /**
      * The logger of this class
@@ -118,6 +119,7 @@ public class TranscriptionGatewaySession
         // We got invited to a room, ready up the transcriber!
         transcriber = new Transcriber(getJvbRoomName(), service);
         transcriber.addTranscriptionListener(this);
+        transcriber.addTranslationListener(this);
         logger.debug("Invited for conference");
     }
 
@@ -284,6 +286,17 @@ public class TranscriptionGatewaySession
     public void notify(TranscriptionResult result)
     {
         sendTranscriptionResultToRoom(result);
+    }
+
+    /**
+     * This method will be called by the {@link TranslationManager} every time a
+     * final transcription result is translated.
+     *
+     * @param result the translation result which has come in
+     */
+    @Override
+    public void notify(TranslationResult result) {
+        sendTranslationResultToRoom(result);
     }
 
     /**
@@ -514,6 +527,16 @@ public class TranscriptionGatewaySession
     private void sendTranscriptionResultToRoom(TranscriptionResult result)
     {
         handler.publishTranscriptionResult(this.chatRoom, result);
+    }
+
+    /**
+     * Send a {@link TranslationResult} to the {@link ChatRoom}
+     *
+     * @param result the {@link TranscriptionResult} to send
+     */
+    private void sendTranslationResultToRoom(TranslationResult result)
+    {
+        handler.publishTranslationResult(this.chatRoom, result);
     }
 
     @Override
--- a/src/main/java/org/jitsi/jigasi/TranscriptionGatewaySession.java
+++ b/src/main/java/org/jitsi/jigasi/TranscriptionGatewaySession.java
@@ -18,6 +18,7 @@
 package org.jitsi.jigasi;
 
 import net.java.sip.communicator.impl.protocol.jabber.*;
+import net.java.sip.communicator.impl.protocol.jabber.extensions.jitsimeet.*;
 import net.java.sip.communicator.service.protocol.*;
 import net.java.sip.communicator.service.protocol.event.*;
 import net.java.sip.communicator.service.protocol.media.*;
@@ -26,6 +27,7 @@ import org.jitsi.jigasi.xmpp.*;
 import org.jitsi.service.neomedia.*;
 import org.jitsi.service.neomedia.device.*;
 import org.jitsi.util.*;
+import org.jivesoftware.smack.packet.*;
 import org.jxmpp.jid.*;
 import org.jxmpp.jid.impl.*;
 import org.jxmpp.stringprep.*;
@@ -257,6 +259,24 @@ public class TranscriptionGatewaySession
 
         String identifier = getParticipantIdentifier(chatMember);
         this.transcriber.participantLeft(identifier);
+    }
+
+    @Override
+    void notifyChatRoomMemberUpdated(ChatRoomMember chatMember, Presence presence)
+    {
+        super.notifyChatRoomMemberUpdated(chatMember, presence);
+
+        TranslationLanguageExtension translationLanguage
+            = presence.getExtension(
+                TranslationLanguageExtension.ELEMENT_NAME,
+                TranslationLanguageExtension.NAMESPACE);
+
+        if(translationLanguage != null) {
+            String identifier = getParticipantIdentifier(chatMember);
+            String language = translationLanguage.getTranslationLanguage();
+
+            this.transcriber.updateParticipantLanguage(identifier, language);
+        }
     }
 
     @Override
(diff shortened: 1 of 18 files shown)


src/main/java/org/jitsi/jigasi/TranscriptionGatewaySession.java
```diff
@@ -26,6 +27,7 @@
 import org.jitsi.service.neomedia.*;
 import org.jitsi.service.neomedia.device.*;
 import org.jitsi.util.*;
+import org.jivesoftware.smack.packet.*;
```
pvgupta24: Should be updated to `import org.jivesoftware.smack.packet.Presence;` as 
`org.jivesoftware.smack.packet.Message` &  `net.java.sip.communicator.service.protocol.Message` results in ambiguity.

src/main/java/org/jitsi/jigasi/TranscriptionGatewaySession.java
```diff
@@ -259,6 +261,24 @@ void notifyChatRoomMemberLeft(ChatRoomMember chatMember)
...
+            = presence.getExtension(
+                TranslationLanguageExtension.ELEMENT_NAME,
+                TranslationLanguageExtension.NAMESPACE);
+
+        if(translationLanguage != null) {
+            String identifier = getParticipantIdentifier(chatMember);
+            String language = translationLanguage.getTranslationLanguage();
+
+            this.transcriber.updateParticipantLanguage(identifier, language);
+        }
```
bgrozev: We need to handle the case where the user removes a language (by removing the extension from presence)

src/main/java/org/jitsi/jigasi/transcription/Transcriber.java
```diff
@@ -284,6 +307,28 @@ public void updateParticipant(String identifier,
...
+     * @param identifier the identifier of the participant
+     * @param language the language tag to be updated for the participant
+     */
+    public void updateParticipantLanguage(String identifier, String language)
+    {
+        Participant participant = getParticipant(identifier);
+
+        if(participant != null)
+        {
+            String previousTargetLanguage = participant.getTranslationLanguage();
```
nikvaessen: lines should not exceed 80 characters

src/main/java/org/jitsi/jigasi/transcription/LocalTxtTranscriptHandler.java
```diff
@@ -195,6 +195,17 @@ public void publish(ChatRoom chatRoom, TranscriptionResult result)
...
 
+    /**
+     * Translated Speech-to-text results can be published to the Chatroom
+     * as plain-text here.
+     *
+     * {@inheritDoc}
+     */
+    @Override
+    public void publish(ChatRoom chatRoom, TranslationResult result)
+    {
```
nikvaessen: maybe add a comment why we leave this empty

src/main/java/org/jitsi/jigasi/transcription/LocalJsonTranscriptHandler.java
```diff
@@ -205,10 +205,18 @@
     public final static String JSON_KEY_TYPE = "type";
 
     /**
-     * This field stores the value of the type of the muc message as a string
+     * This field stores the value of the type of the muc message  for
+     * a transcription result to be sent.
      */
     public final static String JSON_VALUE_TYPE = "transcription-result";
```
nikvaessen: to be consistent, we could rename this variable to something like JSON_VALUE_TYPE_TRANSCRIPTION_RESULT

src/main/java/org/jitsi/jigasi/transcription/LocalJsonTranscriptHandler.java
```diff
@@ -242,6 +258,29 @@ public static JSONObject createJSONObject(TranscriptionResult result)
...
     }
 
+    /**
+     *Creates a json object representing the <tt>TranslationResult</tt>.
+     *
+     * @param result the object to be used to produce json.
+     * @return json object representing the <tt>TranslationResult</tt>.
+     */
+    @SuppressWarnings("unchecked")
+    private static JSONObject createTranslationJSONObject(TranslationResult result)
```
nikvaessen: maybe also rename `createJSONObject` to createTranscriptionJSONObject`

src/main/java/org/jitsi/jigasi/transcription/TranslationManager.java
```diff
@@ -0,0 +1,160 @@
...
+ * @author Praveen Kumar Gupta
+ */
+public class TranslationManager
+    implements TranscriptionListener
+{
+
+    /**
+     * List of target languages for translating the transcriptions.
+     */
+    private Map<String, Integer> languages = new HashMap<>();
```
nikvaessen: I don't think this HashMap is threadsafe. Can you make it so calls to the map are synchronised?

src/main/java/org/jitsi/jigasi/transcription/TranslationManager.java
```diff
@@ -0,0 +1,160 @@
...
+     *
+     * @param result the TranscriptionResult notified to the TranslationManager
+     * @return list of TranslationResults
+     */
+    public ArrayList<TranslationResult> getTranslations(
+        TranscriptionResult result)
+    {
+        ArrayList<TranslationResult> translatedResults
+            = new ArrayList<>();
+
```
nikvaessen: when you make this threadsafe, consider only synchronising making a copy so that we don't block the map for a long period of time.

src/main/java/org/jitsi/jigasi/transcription/TranslationManager.java
```diff
@@ -0,0 +1,160 @@
...
+        for(String language : languages.keySet())
+        {
+            String translatedText = translationService.translate(
+                result.getAlternatives().iterator().next().getTranscription(),
```
nikvaessen: Can we check if the iterator actually has a result? if not, this might lead to an exception.

src/main/java/org/jitsi/jigasi/transcription/TranslationManager.java
```diff
@@ -0,0 +1,160 @@
...
+                result.getLanguage().substring(0,2),
```
nikvaessen: why are we taking the substring?

src/main/java/org/jitsi/jigasi/transcription/TranslationManager.java
```diff
@@ -0,0 +1,160 @@
...
+                language);
+
+            translatedResults.add(new TranslationResult(
+                result,
+                language,
+                translatedText));
+        }
+        return translatedResults;
```
nikvaessen: I prefer a newline before returning :)

src/main/java/org/jitsi/jigasi/transcription/TranslationManager.java
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    /**
+     * Translates the received {@link TranscriptionResult} into required languages
+     * and returns a list of {@link TranslationResult}s.
+     *
+     * @param result the TranscriptionResult notified to the TranslationManager
+     * @return list of TranslationResults
+     */
+    public ArrayList<TranslationResult> getTranslations(
```
nikvaessen: should this method be public?

nikvaessen: Make this return a `List` instead of `ArrayList`. Internally, we can still return an `ArrayList`.

src/main/java/org/jitsi/jigasi/transcription/TranslationManager.java
```diff
@@ -0,0 +1,160 @@
...
+    @Override
+    public void notify(TranscriptionResult result)
+    {
+        if(!result.isInterim())
+        {
+            ArrayList<TranslationResult> translations
+                = getTranslations(result);
+            for (TranslationResultListener listener : listeners)
+            {
+                for (TranslationResult translation : translations)
```
nikvaessen: ` translations.forEach(listener::notify);` might be more elegant. Or even
`listeners.forEach(l -> translations.forEach(l::notify));`

src/main/java/org/jitsi/jigasi/transcription/TranslationManager.java
pvgupta24: No.

src/main/java/org/jitsi/jigasi/transcription/TranslationManager.java
pvgupta24: Google CloudTranslate expects only the language code without country code.
I will update it to pass the entire language-code and take the language code in `GoogleCloudTranslationService.java`

PR #129, opened 2018-08-10
make transcriber leave a conference based on participants presence

nikvaessen:
While Jicofo will invite a transcriber, Jigasi itself will decide when it's time to leave.


nikvaessen:
update jitsi-desktop-version before merging

bgrozev:
Pending the use of a new property instead of 'source language', as discussed offline. Otherwise, LGTM

PR #130, opened 2018-08-12
WIP : Saves final transcript in translated languages

pvgupta24:


pvgupta24:
@nikvaessen @bgrozev PTAL.

PR #134, opened 2018-08-27
catch "deallocate cannot be used on a started conference" exceptions

nikvaessen:
This will make Jigasi leave the conference, but the session needs to be restarted in order for a reinvite to appear as jicofo will not be told that TranscriptionStatus = off, as the exception causes updating presence to fail.


bgrozev:
This required a rebase so I opened #137

PR #136, opened 2018-08-28
Few improvements.

damencho:
Few fixes for exceptions I saw during testing.

Fixes memory leak, leaking and not disconnecting bosh connections. I still see the nginx server timeout, but it is very rear and I still see it on the muc control connection, which reconnects maybe some timing need to be tuned.

I was reproducing 10-20% of the times not working transcription, moving the setConference call earlier (before answering the call) fixes it for me.

Sometimes I was seeing the conference to timeout, and this was because kick of the transcription participant was not detected in the code. Sometimes the listener for the kick is invoked and sometimes the participant leave one by one and we were not detecting correctly that the focus has left.

### DIFF
--- a/src/main/java/org/jitsi/jigasi/JvbConference.java
+++ b/src/main/java/org/jitsi/jigasi/JvbConference.java
@@ -830,7 +830,7 @@ public class JvbConference
         }
 
         // if it is the focus leaving
-        if (member.getContactAddress().equals(focusResourceAddr))
+        if (member.getName().equals(focusResourceAddr))
         {
             logger.info("Focus left! - stopping");
             stop();
@@ -861,7 +861,6 @@ public class JvbConference
     public void localUserPresenceChanged(
         LocalUserChatRoomPresenceChangeEvent evt)
     {
-
         if (evt.getChatRoom().equals(JvbConference.this.mucRoom)
             && Objects.equals(evt.getEventType(),
                     LocalUserChatRoomPresenceChangeEvent.LOCAL_USER_KICKED))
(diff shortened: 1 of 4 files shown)


src/main/java/org/jitsi/jigasi/JvbConference.java
```diff
@@ -1332,13 +1333,26 @@ private void inviteFocus(final EntityBareJid roomIdentifier)
...
+        /**
+         * Whether the timeout had been scheduled.
+         * @return <tt>true</tt> if timeout had been scheduled,
+         * <tt>false</tt> otherwise.
+         */
+        boolean isScheduled()
+        {
+            synchronized (syncRoot)
+            {
+                return timeoutThread != null;
```
bgrozev: Why do we have a synchronized block just for a null check?

damencho: So when the transcriber is kicked we can get two MEMBER_LEFT events on memberPresenceChanged, very quickly one after another, as I'm not sure are those running in the same thread or not ... but I was seeing IllegalStateException several times (the exception thrown after isScheduled() in the code below).
The first can start creating the timeoutThread and we don't want the second to try this at the same time and get IllegalStateException... just to wait for the other to finish work before returning the result in memberPresenceChanged
```
if (!leaveTimeout.isScheduled())
                    leaveTimeout.scheduleTimeout(leaveTime);
```. Just protecting reading of that field while writing in it...
Does this make sense? Or you think we should change it somehow and keep the check near the writing of timeoutThread.

bgrozev: I don't understand how this synchronized block around the null check fixes the problem that you describe. If you have two threads execute the `if`, the race is still possible. It seems to me that you need to synchronize the whole `if`.

damencho: Ups, now I see what you mean :)

damencho: Does it look better now?

PR #137, opened 2018-08-29
catch "deallocate cannot be used on a started conference" exceptions

bgrozev:
Just a rebased version of 134


damencho:
This had been fixed at some point, it was something about the time of merging the calls ...

PR #138, opened 2018-08-30
Fixes graceful shutdown and report it in stats.

damencho:


src/main/java/org/jitsi/jigasi/JigasiBundleActivator.java
```diff
@@ -235,4 +265,72 @@ public void serviceChanged(ServiceEvent serviceEvent)
             sipGateway.setSipProvider(pps);
         }
     }
+
```
bgrozev: Statics on top. Also, why is everything static?

PR #140, opened 2018-09-05
sip health checking

damencho:


src/main/java/org/jitsi/jigasi/health/Health.java
```diff
@@ -0,0 +1,131 @@
...
+            }
+            else
+            {
+                // we will wait for the sip gw to become ready and will try
+                // to start it again
+                final SipGateway finalSipGw = sipGateway;
+                sipGateway.addGatewayListener(new GatewayListener()
+                {
+                    @Override
+                    public void onSessionAdded(AbstractGatewaySession session)
```
bgrozev: Can we add default implementations in the GatwayListener interface?

src/main/java/org/jitsi/jigasi/xmpp/CallControlMucActivator.java
```diff
@@ -344,6 +344,10 @@ public void onSessionRemoved(AbstractGatewaySession session)
     public void onSessionFailed(AbstractGatewaySession session)
     {}
 
+    @Override
+    public void onReady()
```
bgrozev: We already use java8 in jigasi, let's just get rid of all these dummy overrides and use default methods in the interface

PR #142, opened 2018-11-01
Enable the use of better performing "video model" of Google Speech API

nikvaessen:
Good news! The video model of the Google Cloud speech-to-text API can now be used. While it's more expensive to use, it also offers better transcription capabilities.

When this model was initially released in May 2018 it was not able to return IS_FINAL results. However, after trying to implement it again now it turns it is does return IS_FINAL results, making it possible to close the current StreamingSession and open a new one without having to do VAD ourselves.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -27,17 +27,17 @@
     <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
-      <version>23.0</version>
+      <version>27.0-jre</version>
     </dependency>
     <dependency>
       <groupId>com.google.cloud</groupId>
       <artifactId>google-cloud-speech</artifactId>
-      <version>0.49.0-alpha</version>
+      <version>0.68.0-beta</version>
     </dependency>
     <dependency>
       <groupId>com.google.cloud</groupId>
       <artifactId>google-cloud-translate</artifactId>
-      <version>1.31.0</version>
+      <version>1.50.0</version>
     </dependency>
     <dependency>
       <groupId>dom4j</groupId>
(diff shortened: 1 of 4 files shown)


src/main/java/org/jitsi/jigasi/transcription/GoogleCloudTranscriptionService.java
```diff
@@ -179,12 +179,52 @@
...
+     */
+    private final static String P_NAME_USE_VIDEO_MODEL
+        = "org.jitsi.jigasi.transcription.USE_VIDEO_MODEL";
+
+    /**
+     * The default value for the property USE_VIDEO_MODEL
+     */
+    private final static boolean DEFAULT_VALUE_USE_VIDEO_MODEL = false;
+
+    /**
```
nikvaessen: I moved this up because I removed some static methods

src/main/java/org/jitsi/jigasi/transcription/GoogleCloudTranscriptionService.java
```diff
@@ -179,12 +179,52 @@
...
+     * Check whether the given string contains a supported language tag
+     *
+     * @param tag the language tag
+     * @throws UnsupportedOperationException when the google cloud API does not
+     * support the given language
+     */
+    private static void validateLanguageTag(String tag)
+        throws UnsupportedOperationException
```
nikvaessen: I moved this up because I made some methods which were above non-static, but want to preserve static methods being at top of file.

pom.xml
```diff
@@ -27,17 +27,17 @@
...
       <groupId>com.google.cloud</groupId>
       <artifactId>google-cloud-speech</artifactId>
-      <version>0.49.0-alpha</version>
+      <version>0.68.0-beta</version>
     </dependency>
     <dependency>
       <groupId>com.google.cloud</groupId>
       <artifactId>google-cloud-translate</artifactId>
-      <version>1.31.0</version>
+      <version>1.50.0</version>
```
nikvaessen: The google-cloud-translate API needs to be updated because of dependency issues; not updating it made it impossible to receive results from the Google Cloud server.

PR #190, opened 2019-08-20
Adapts to changes in jitsi-stats.

bgrozev:
Adapts to changes in jitsi/jitsi-stats#8


damencho:
But shouldn't we change here and the version in the pom file?

bgrozev:
Yes, I will update it once we merge the jitsi-stats PR

PR #192, opened 2019-08-21
Sip no hangup on jingle terminate

damencho:


src/main/java/org/jitsi/jigasi/JvbConference.java
```diff
@@ -285,50 +285,6 @@ public JvbConference(AbstractGatewaySession gatewaySession, CallContext ctx)
             .getString(FOCUSE_RESOURCE_PROP, "focus");
     }
 
-    /**
-     * Includes info about given <tt>peer</tt> media SSRCs in MUC presence.
-     * @param peer the <tt>CallPeer</tt> whose media SSRCs will be advertised.
-     */
-    private void advertisePeerSSRCs(CallPeer peer)
```
bgrozev: Is this just unnecessary since nothing reads SSRCs from precense?

damencho: Yep, I searched around the namespace and didn't found anything ...

src/main/java/org/jitsi/jigasi/SipGatewaySession.java
```diff
@@ -407,6 +407,16 @@ public void peerStateChanged(CallPeerChangeEvent evt)
                 });
             }
         }
+
+        logger.info("JVB conference call IN_PROGRESS "
```
bgrozev: I don't understand what this means, even after looking at the source code. `IN_PROGRESS` seems to refer to a `CallState`, but is this necessarily the state of the call? Please, let's improve the log message.

src/main/java/org/jitsi/jigasi/SipCallTransformerMonitor.java
```diff
@@ -0,0 +1,228 @@
...
+        EXECUTOR.registerRecurringRunnable(recurringMediaChecker);
+    }
+
+    /**
+     * Disposes, clears the executor and its task.
+     */
+    void dispose()
+    {
+        EXECUTOR.deRegisterRecurringRunnable(recurringMediaChecker);
+        EXECUTOR.close();
```
bgrozev: Do we really want to close the static executor when we dispose of a single instance?

damencho: Oh :) will fix it.

src/main/java/org/jitsi/jigasi/SipCallTransformerMonitor.java
```diff
@@ -0,0 +1,228 @@
...
+
+    /**
+     * The RTCP transformer.
+     */
+    private final RTCPTransformer rtcpTransformer = new RTCPTransformer();
+
+    /**
+     * List of seen incoming or outgoing ssrcs.
+     */
+    private Set<Long> seenSSRCs = new HashSet<>();
```
bgrozev: The transformer chain is run in different threads, so this needs to be thread safe. We can use `ConcurrentHashMap.newKeySet()` for example

src/main/java/org/jitsi/jigasi/SipGatewaySession.java
damencho: Yep, will improve it, It was left like that while juggling with the code.

src/main/java/org/jitsi/jigasi/SipCallTransformerMonitor.java
```diff
@@ -0,0 +1,228 @@
...
+        public void run()
+        {
+            super.run();
+
+            try
+            {
+                if (System.currentTimeMillis()
+                        - lastOutgoingActivity > getPeriod())
+                {
+                    handler.sendHolePunchPacket(stream, MediaType.AUDIO);
```
bgrozev: I see a few problems or potential problems here.
1. Hole-punch packets are meant for the start of a session. Looking at the jitsi code that sends them, they will have a valid PT and SSRC, but they have a hardcoded sequence number (0x29e). Sending such a packet in the middle of a stream could lead to problems (e.g. it could brake the SRTP session). 

2. We're sending a single packet, and it could be lost.

3. The timeout period is the same as period of the executable -- 25 seconds. This means that in some cases we'll start sending keep alive packets up to 50 seconds after the last activity. IIRC the threshold we're after is 35 seconds.

Is there a standarsized or widely-used approach for sending keep alive packets with RTP? [This document](https://tools.ietf.org/html/rfc6263#section-5) recommends using RTCP, e.g. we could send empty RTCP RRs (or something else?) instead.

src/main/java/org/jitsi/jigasi/SipCallTransformerMonitor.java
```diff
@@ -0,0 +1,228 @@
...
+ * Monitors outgoing RTCP and passes only those which are relevant for that
+ * direction. Filters outgoing RTCP.BYEs for ssrc that we haven't seen. Also
+ * monitors last seen media or rtcp and if one is missing we
+ * send hole punch packets. It can happen that the xmpp call is dropped and
+ * the sip call participant is waiting then there will be no media ot RTCP to
+ * keep alive the stream from the other side.
+ *
+ * @author Damian Minkov
+ */
+public class SipCallTransformerMonitor
```
bgrozev: Perhaps `SipCallKeepAliveTransformer` might be a better name.

damencho: I chose monitor because it filters and RTCP :) but I'll change it

src/main/java/org/jitsi/jigasi/SipCallTransformerMonitor.java
damencho: Talked offline and will try implementing this with different ssrc and RTP and PT 13(comfort noise).

PR #195, opened 2019-08-26
Add/remove XMPP MUCs using a REST endpoint.

damencho:


src/main/java/org/jitsi/jigasi/xmpp/CallControlMucActivator.java
```diff
@@ -268,15 +272,13 @@ public void registrationStateChanged(RegistrationStateChangeEvent evt)
...
-        OperationSetMultiUserChat muc
-            = pps.getOperationSet(OperationSetMultiUserChat.class);
-
         String roomName = pps.getAccountID()
             .getAccountPropertyString(ROOM_NAME_ACCOUNT_PROP);
 
         if (roomName == null)
         {
-            logger.warn("No brewery name specified for:" + pps);
+            // this is the common account for connecting which is configured
```
bgrozev: How do we know that this is the account which is not supposed to have a brewery? Wouldn't this happen if one of the other accounts in misconfigured, in which case it is helpful to have the log message? Why does the comment mention authorization, is it related to this at all?

damencho: Hum, wrong comment, will fix it :( 
So this is a RegistrationStateChangeListener which we add to the protocol provider ... hum, I will further check it ... I was seeing this log "No brewery name specified for:" multiple times in production ... and wanted to remove it as it is not relevant ... but you make me think that something else is happening and I will check it ...

src/main/java/org/jitsi/jigasi/rest/HandlerImpl.java
```diff
@@ -325,4 +358,113 @@ static synchronized void sendJSON(
...
+                    CallControlMucActivator.BREWERY_ENABLED_PROP, false))
+        {
+            //MUC call control disabled.
+            response.setStatus(HttpServletResponse.SC_SERVICE_UNAVAILABLE);
+            return;
+        }
+
+        if ("add".equals(target))
+        {
+            if (requestJSONObject == null
```
bgrozev: There's already a null check above

src/main/java/org/jitsi/jigasi/rest/HandlerImpl.java
```diff
@@ -325,4 +358,113 @@ static synchronized void sendJSON(
...
+                response.setStatus(
+                    HttpServletResponse.SC_INTERNAL_SERVER_ERROR);
+                return;
+            }
+
+            response.setStatus(HttpServletResponse.SC_OK);
+        }
+        else if ("remove".equals(target))
+        {
+            if (requestJSONObject == null
```
bgrozev: If both valid cases require the "id" to be present, why not read it once?

src/main/java/org/jitsi/jigasi/xmpp/CallControlMucActivator.java
```diff
@@ -400,6 +401,101 @@ private void updatePresenceStatusForXmppProviders()
...
+    /**
+     * Adds new call control MUC xmpp account using the provided props.
+     *
+     * @param id the id to use for this account(should not contain .).
+     * @param propertiesSet the properties.
+     * @throws OperationFailedException if failed loading account.
+     */
+    public static void addCallControlMucAccount(
+            String id,
+            Set<Map.Entry<String, String>> propertiesSet)
```
bgrozev: Why deal with a set of entries instead of the map itself?

src/main/java/org/jitsi/jigasi/xmpp/CallControlMucActivator.java
```diff
@@ -400,6 +401,101 @@ private void updatePresenceStatusForXmppProviders()
...
+        AccountManager accountManager
+            = ProtocolProviderActivator.getAccountManager();
+
+        // extract the properties, make sure we decode the password
+        HashMap<String, String> properties = new HashMap<>();
+        propertiesSet.stream().forEach(e ->
+        {
+            String property = e.getKey();
+            String value = e.getValue();
+            if (ProtocolProviderFactory.PASSWORD.equals(property))
```
bgrozev: Why do we expect that the pasword is Base64 encoded?

src/main/java/org/jitsi/jigasi/xmpp/CallControlMucActivator.java
```diff
@@ -400,6 +401,101 @@ private void updatePresenceStatusForXmppProviders()
...
+        if (accountID != null)
+        {
+            boolean result
+                = xmppProviderFactory.uninstallAccount(accountID);
+            logger.info("Removing muc control account: "
+                + id + ", successful:" + result);
+        }
+        else
+        {
+            logger.warn("No muc control account found for removing");
```
bgrozev: We should propagate this to HTTP

src/main/java/org/jitsi/jigasi/xmpp/CallControlMucActivator.java
```diff
@@ -400,6 +401,101 @@ private void updatePresenceStatusForXmppProviders()
...
+            = xmppProviderFactory.createAccount(properties);
+
+        // A small workaround to make sure we use the id for the
+        // account that is provided, this is in case we want to
+        // delete the account by the same id.
+        // We use the fact that account manager will reuse the id if found
+        // to store the properties (like edit of an account).
+        {
+            String accountConfigPrefix = accountManager
+                .getFactoryImplPackageName(xmppProviderFactory) + "." + id;
```
bgrozev: If all of the other properties will use the full jitsi format, why do we treat "id" in a special way? Can't we just add the prefix in the HTTP request?

src/main/java/org/jitsi/jigasi/xmpp/CallControlMucActivator.java
damencho: This is how we configure it in the sip-communicator.properties file, I thought to keep it ... as to be able to use the same template for adding those accounts ... 
And I tested, removing the XMPP account from my config and started adding same properties and values using the REST endpoint ...

src/main/java/org/jitsi/jigasi/xmpp/CallControlMucActivator.java
damencho: Well ... I kept the same API as in the jvb, as described in the doc. If json is correct 200 OK is returned ... no further output as a result ... but if you say I will change it ...

PR #200, opened 2019-09-26
Passes all header params from rayo to sip and xmpp side.

damencho:


src/main/java/org/jitsi/jigasi/SipGatewaySession.java
```diff
@@ -509,12 +509,30 @@ public void outgoingCallCreated(CallEvent callEvent)
                     if(roomName != null)
                     {
                         Call call = callEvent.getSourceCall();
-                        call.setData("EXTRA_HEADER_NAME.1",
+                        AtomicInteger headerCount = new AtomicInteger(0);
```
gpolitis: It doesn't hurt to use an atomic int but I'm struggling to understand what's its purpose this particular case 🤔

damencho: This is because in the lambda the variables need to be final or effectively final... And this int is incremented and used there.

src/main/java/org/jitsi/jigasi/SipGatewaySession.java
```diff
@@ -509,12 +509,30 @@ public void outgoingCallCreated(CallEvent callEvent)
...
                                 .getAccountPropertyString(
                                     JITSI_MEET_ROOM_HEADER_PROPERTY,
                                     "Jitsi-Conference-Room"));
-                        call.setData("EXTRA_HEADER_VALUE.1", roomName);
+                        call.setData(
+                            "EXTRA_HEADER_VALUE." + headerCount.get(),
+                            roomName);
+
+                        // passes all extra headers to the outgoing call
+                        callContext.getExtraHeaders().forEach(
```
gpolitis: Can the extra headers map be modified by another thread while we iterate over them? I see we return the raw object. Would it make sense to return a read only map?

src/main/java/org/jitsi/jigasi/SipGatewaySession.java
gpolitis: got it, thanks damencho

src/main/java/org/jitsi/jigasi/SipGatewaySession.java
gpolitis: I'm good with the code as it is, maybe you could just add your comment as an XXX in the code so that we can remember this reasoning if and when we decide to touch this again.

PR #203, opened 2019-10-30
Detects maximum occupants limit reached to play an announcement.

damencho:


src/main/java/org/jitsi/jigasi/SipGatewaySession.java
```diff
@@ -162,6 +163,24 @@
...
+
+    /**
+     * Approximate duration of the file to be played, we need it as to know
+     * when to hangup the call. The actual file is 10 seconds but we give a
+     * little longer for the file to be played.
+     */
+    private static final int MAX_OCCUPANTS_SOUND_DURATION_SEC = 15;
+
+    /**
+     * In certain scenarios (mac occupants) we wait till we hangup the call.
```
gpolitis: max*

src/main/java/org/jitsi/jigasi/JvbConference.java
```diff
@@ -746,6 +748,15 @@ private void joinConferenceRoom()
         }
         catch (Exception e)
         {
+            if (e.getCause() instanceof XMPPException.XMPPErrorException)
+            {
+                if (((XMPPException.XMPPErrorException)e.getCause())
+                        .getXMPPError().getCondition() == service_unavailable)
+                {
+                    gatewaySession.notifyMaxOccupantsLimitReached();
```
gpolitis: Non-blocking question/comment. Notifications are (at least in my mind..) handled quickly, but it seems here we block in order to handle the failure. If we want to handle the case where we reach the max occupants, why not just name the method `handleMaxOccupantsLimitReached`? Is it for consistency with the rest of the  method names?

PR #219, opened 2020-01-09
Post debug extras

paweldomas:
This is two extra commits added on top of https://github.com/jitsi/jigasi/pull/218 but it's a big change and not sure if desired.


src/main/java/org/jitsi/jigasi/transcription/GoogleCloudStreamingRecognitionSession.java
```diff
@@ -0,0 +1,692 @@
...
+     * audio in "real-time". This also ensures that participants using push-
+     * to-talk do not have delayed results
+     */
+    private final static int STREAMING_SESSION_TIMEOUT_MS = 2000;
+
+    /**
+     * The logger for this class
+     */
+    private final static Logger logger
+        = Logger.getLogger(GoogleCloudTranscriptionService.class);
```
nikvaessen: Change the logger name?

src/main/java/org/jitsi/jigasi/transcription/GoogleCloudStreamingRecognitionSession.java
```diff
@@ -0,0 +1,692 @@
...
+        public void onError(Throwable t)
+        {
+            logger.warn(debugName + ": received an error from the Google Cloud API", t);
+            requestManager.terminateCurrentSession();
+        }
+
+        @Override
+        public void onCompleted()
+        {
+            for(TranscriptionListener listener : requestManager.getListeners())
```
nikvaessen: I think it would be valuable to trace what listeners have been to stop or whether whether the list of listeners is empty

PR #222, opened 2020-01-16
Reconnect update

damencho:
This PR's goal was to minimize the synchronization in the reconnect logic, that we saw several times jigasi filing incoming calls because of some lock in the reconnect logic ... 
Needs updating the pom.xml after merging: https://github.com/jitsi/jitsi/pull/636 as this is the main change.
While testing several scenarios of reconnecting (restarting prosody and stopping in/out on port 443 on the server with iptables) fixed few other bugs like authenticated dynamic accounts not able reconnect.

### DIFF
--- a/src/main/java/org/jitsi/jigasi/JvbConference.java
+++ b/src/main/java/org/jitsi/jigasi/JvbConference.java
@@ -255,6 +255,11 @@ public class JvbConference
     private String jvbParticipantStatus = null;
 
     /**
+     * Synchronizes the write access to {@code jvbParticipantStatus}.
+     */
+    private final Object statusSync = new Object();
+
+    /**
      * Call hang up reason string that will be sent to the SIP peer.
      */
     private String endReason;
@@ -789,11 +794,16 @@ public class JvbConference
         }
     }
 
-    synchronized void setPresenceStatus(String statusMsg)
+    void setPresenceStatus(String statusMsg)
     {
-        if (statusMsg.equals(jvbParticipantStatus))
+        synchronized(statusSync)
         {
-            return;
+            if (statusMsg.equals(jvbParticipantStatus))
+            {
+                return;
+            }
+
+            jvbParticipantStatus = statusMsg;
         }
 
         if (mucRoom != null)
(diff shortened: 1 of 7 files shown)


src/main/java/org/jitsi/jigasi/JvbConference.java
```diff
@@ -1546,4 +1562,57 @@ public String toString()
...
     }
+
+    /**
+     * Extracts bosh connection sessionId, we just print it for debug
+     * purposes.
+     * @param connection the bosh connection which sessionId we will try to
+     * extract.
+     * @return the sessionId if extracted or null.
+     */
+    private static Object getConnSessionId(Object connection)
```
paweldomas: It's not a big deal, but I would rather avoid putting utility methods in relatively large already `JvbConference` class.

src/main/java/org/jitsi/jigasi/JvbConference.java
```diff
@@ -784,8 +789,13 @@ private void joinConferenceRoom()
         }
     }
 
-    void setPresenceStatus(String statusMsg)
+    synchronized void setPresenceStatus(String statusMsg)
```
paweldomas: I'm wondering if maybe `synchronized` should only cover the `if` condition and the state modification. I suppose that `jitsiMeetTools.setPresenceStatus(` will send all changes in order and with proper synchronization. This is in the context of knowing that sending XMPP update may block for a while and make even the NO-OP threads(`statusMsg.equals(jvbParticipantStatus)`) wait for it to finish.

This is what I mean:
```
synchronized(this) {
  if (statusMsg.equals(jvbParticipantStatus))
        {
            return;
        }
  jvbParticipantStatus = statusMsg;
}


if (mucRoom != null)
{
            // Send presence status update
            OperationSetJitsiMeetTools jitsiMeetTools
                = xmppProvider.getOperationSet(
                    OperationSetJitsiMeetTools.class);


            jitsiMeetTools.setPresenceStatus(mucRoom, statusMsg);
}
```

damencho: Done.

src/main/java/org/jitsi/jigasi/JvbConference.java
damencho: Changed.

PR #223, opened 2020-01-17
Stabilize google transcriptions

paweldomas:
Drop unstable results early.


src/main/java/org/jitsi/jigasi/transcription/GoogleCloudTranscriptionService.java
```diff
@@ -955,28 +965,38 @@ private boolean isEndOfSingleUtteranceMessage(
...
-
-            if(alternatives.isEmpty())
+            if (result.getAlternativesList().isEmpty())
             {
+                throw new IllegalArgumentException(
+                        "The alternatives list must not be empty");
+            }
+
+            SpeechRecognitionAlternative alternative = result.getAlternatives(0);
+            String newStablePart = alternative.getTranscript();
```
bgrozev: Why are they called `part` and `stableText`? Perhaps use more descriptive names like `latestTranscript` and `newTranscript`?

PR #232, opened 2020-02-14
Introduces sound notification manager.

damencho:
This PR should not change any logic, just extracts code dealing with the sound notifications played to the sip participant for recordings/live-streaming and max occupants and moves it in a separate class, so we can later add even more notifications and simplify the code in JvbConference and SipGatewaySession.

### DIFF
--- a/src/main/java/org/jitsi/jigasi/AbstractGatewaySession.java
+++ b/src/main/java/org/jitsi/jigasi/AbstractGatewaySession.java
@@ -67,6 +67,19 @@ public abstract class AbstractGatewaySession
     protected boolean gatewayMediaDropped = false;
 
     /**
+     * Configuration property to change the resource used by focus.
+     */
+    private static final String FOCUSE_RESOURCE_PROP
+        = "org.jitsi.jigasi.FOCUS_RESOURCE";
+
+    /**
+     * Address of the focus member that has invited us to the conference.
+     * Used to identify the focus user and dispose the session when it leaves
+     * the room.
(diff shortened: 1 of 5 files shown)


src/main/java/org/jitsi/jigasi/AbstractGatewaySession.java
```diff
@@ -66,6 +66,19 @@
...
+     */
+    private static final String FOCUSE_RESOURCE_PROP
+        = "org.jitsi.jigasi.FOCUS_RESOURCE";
+
+    /**
+     * Address of the focus member that has invited us to the conference.
+     * Used to identify the focus user and dispose the session when it leaves
+     * the room.
+     */
+    protected final String focusResourceAddr;
```
paweldomas: can you add getter/setter for this property instead of accessing directly in other classes?

paweldomas: I don't really see any simplification coming out of this move. Do you plan to use different focus resources on per session basis? The configuration option is global, which means this property is static and not bound to any specific instance. Maybe it should be a static variable in one of the bundle activators?

damencho: I moved it there cause I needed a check moved in SoundNotificationManager.
https://github.com/jitsi/jigasi/pull/232/files#diff-d1aca1f84c48e995cd1e763c8d36828fR131
I will add the getters.

PR #236, opened 2020-03-03
Add startmuted support and sip unmute

theunafraid:


src/main/java/org/jitsi/jigasi/JvbConference.java
```diff
@@ -1599,4 +1609,111 @@ public String toString()
...
+            MuteIq muteIq = new MuteIq();
+            muteIq.setJid(memberJid);
+            muteIq.setMute(bMuted);
+            muteIq.setType(IQ.Type.set);
+            muteIq.setTo(roomJid);
+            // From is ignored, XMPPConnection changes the from unless
+            // FromMode is set to UNCHANGED, Actor can be used instead
+            muteIq.setFrom(fromJid);
+            muteIq.setActor(fromJid);
+
```
damencho: In general, you don't need to set from. It is a responsibility of the xmpp server to set it when sending to other xmpp endpoints. This is because of security, you can spoof with different from, but the server knows the authentication details on the current connection and set them before sending the packets to other participants.
The actor also should not be mandatory. We added that to be able to say who muted someone else, but when you are muting yourself this is not important.
Have you tested it with removing these two lines and also removing the 
```
private static final String JITSI_MEET_JIGASI_USER_ID
        = "org.jitsi.jigasi.xmpp.acc.USER_ID";
```
?

src/main/java/org/jitsi/jigasi/JvbConference.java
```diff
@@ -1599,4 +1609,111 @@ public String toString()
...
+            MuteIq muteIq = new MuteIq();
+            muteIq.setJid(memberJid);
+            muteIq.setMute(bMuted);
+            muteIq.setType(IQ.Type.set);
+            muteIq.setTo(roomJid);
+            // From is ignored, XMPPConnection changes the from unless
+            // FromMode is set to UNCHANGED, Actor can be used instead
+            muteIq.setFrom(fromJid);
+            muteIq.setActor(fromJid);
```
damencho: In general, you don't need to set from. It is a responsibility of the xmpp server to set it when sending to other xmpp endpoints. This is because of security, you can spoof with different from, but the server knows the authentication details on the current connection and set them before sending the packets to other participants.
The actor also should not be mandatory. We added that to be able to say who muted someone else, but when you are muting yourself this is not important.
Have you tested it with removing these two lines and also removing the
```
private static final String JITSI_MEET_JIGASI_USER_ID
        = "org.jitsi.jigasi.xmpp.acc.USER_ID";
```
?

src/main/java/org/jitsi/jigasi/SipGatewaySession.java
```diff
@@ -276,6 +278,12 @@
      */
     private boolean callMaxOccupantsLimitReached = false;
 
+    /**
+     * startMutedFlags[0] represents the muted status of audio stream.
+     * startMuted[1] represents the muted status of video stream.
+     */
+    private boolean[] startMuted = { false, false };
```
damencho: Do we need to handle the video muted stuff as jigasi is audio only and we care only for audio?

damencho:
Also there are many methods without javadocs, please add javadoc for those briefly explaining what they are supposed to do, the params and return type if any.

src/main/java/org/jitsi/jigasi/JvbConference.java
```diff
@@ -1599,4 +1609,111 @@ public String toString()
...
+                // we check for optional muc service
+                String mucService
+                    = JigasiBundleActivator.getConfigurationService()
+                        .getString("org.jitsi.jigasi.MUC_SERVICE_ADDRESS", null);
+                if (!StringUtils.isNullOrEmpty(mucService))
+                {
+                    roomName = roomName + "@" + mucService;
+                }
+            }
+
```
damencho: There should be an easier way to obtain the room address.

damencho: What about mucRoom.getIdentifier()?

src/main/java/org/jitsi/jigasi/JvbConference.java
```diff
@@ -1599,4 +1609,111 @@ public String toString()
...
+            String jidString = roomName + "/" + callContext
+                                                    .getCallResource()
+                                                    .getLocalpartOrNull()
+                                                    .toString();
```
damencho: Maybe use getResourceIdentifier();

src/main/java/org/jitsi/jigasi/JvbConference.java
theunafraid: Removed, works ok without.

src/main/java/org/jitsi/jigasi/SipGatewaySession.java
theunafraid: For video the flag is ignored but wanted to be consistent with startmuted extension.

src/main/java/org/jitsi/jigasi/JvbConference.java
theunafraid: Changed to getIdentifier.

src/main/java/org/jitsi/jigasi/JvbConference.java
theunafraid: Used getResourceIdentifier.

src/main/java/org/jitsi/jigasi/SipGatewaySession.java
damencho: Yeah, but removing it will significantly simplify code.

damencho:
Also, we need this to be behind a flag and disabled by default. Sorry, I forgot about it when I did the initial review.

theunafraid:
> Also, we need this to be behind a flag and disabled by default. Sorry, I forgot about it when I did the initial review.

Added prop to enable or disable startmuted for sip.

src/main/java/org/jitsi/jigasi/SipGatewaySession.java
theunafraid: Removed the flag for the video stream

src/main/java/org/jitsi/jigasi/SipGatewaySession.java
```diff
@@ -263,6 +271,8 @@ public SipGatewaySession(SipGateway gateway, CallContext callContext)
             = sipProvider.getOperationSet(
                     OperationSetJitsiMeetTools.class);
 
+        jitsiMeetTools.addRequestListener(this);
```
damencho: Hum seems we are already setting this listener

damencho: Why do you need it here?

damencho: We should either remove the other one and make sure we clean up, as this will leak sessions in memory.

damencho:
And please rebase to master. I brought you some conflicts as we needed some changes sooner ... sorry.

theunafraid:
@damencho Created a new pr to resolve conflicts : https://github.com/jitsi/jigasi/pull/244

PR #244, opened 2020-03-19
Add startmuted support and sip unmute

theunafraid:


theunafraid:
@damencho Regarding the addRequestListener, sorry forgot about those, initially i did not handle the CONNECTED and DISCONNECTED state. I removed the additional addRequestListener and also updated JvbConference having all of this in mind.

PR #253, opened 2020-04-03
fix(xmpp): disable stream management and resume

paweldomas:
The app is not expecting for that to happen also there is a known smack
issue which could lead to a deadlock:

https://discourse.igniterealtime.org/t/deadlock-between-the-reader-and-writer-with-stream-management/87442/3


avinashpawara64-ops:
Hi

PR #257, opened 2020-04-09
Fix cancel alone playback on join

theunafraid:


damencho:
Can you give more details? What was the problem you are solving? The problem you saw?

theunafraid:
> Can you give more details? What was the problem you are solving? The problem you saw?

Alone playback is played even if not alone in meeting. Join web then dial-in -> Alone.opus is played after ParticipantJoined.opus is played which shouldn't happen because the user who dialed-in is not alone in meeting.

damencho:
Thanks.

PR #258, opened 2020-04-09
Fix remote muting jigasi participants

damencho:


src/main/java/org/jitsi/jigasi/JvbConference.java
```diff
@@ -1416,8 +1414,7 @@ private void inviteFocus(final EntityBareJid roomIdentifier)
             StanzaCollector collector = null;
             try
             {
-                collector = ((ProtocolProviderServiceJabberImpl) xmppProvider)
-                    .getConnection()
+                collector = getConnection()
```
paweldomas, line 1424: can this fail with NPE here if no connection?

paweldomas: maybe it doesn't  matter I dont know.

damencho: Yep, it can, but I don't see how this is possible to try inviting the focus you will get connected. And I just changed to use the getConnection() without changing the logic.

paweldomas: yes

PR #259, opened 2020-04-12
Update Bouncy Castle to 1.70

Neustradamus:
- https://www.bouncycastle.org/releasenotes.html
- http://www.bouncycastle.org/latest_releases.html


bgrozev:
Does this work correctly with the `bcpkix-jdk15on:jar:1.54` inherited from libjitsi?

Neustradamus:
@bgrozev, @saghul: One PR here.

PR #261, opened 2020-04-14
add ability to filter out speechless audio based on WebRTCVAD library

paweldomas:
Rebased version of https://github.com/jitsi/jigasi/pull/144 with fixes to native lib loading in https://github.com/jitsi/jitsi-webrtc-vad-wrapper/commit/59923d3f9ecf8059f9fe61c17d420f5df538ca08

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -156,6 +156,11 @@
       <version>${jitsi-desktop.version}</version>
     </dependency>
     <dependency>
+      <groupId>org.jitsi</groupId>
+      <artifactId>jitsi-webrtcvadwrapper</artifactId>
+      <version>1.0-SNAPSHOT</version>
+    </dependency>
+    <dependency>
       <groupId>${project.groupId}</groupId>
       <artifactId>jitsi-utils</artifactId>
       <version>1.0-39-gd481c98</version>
@@ -327,6 +332,11 @@
       <artifactId>jitsi-systray-service</artifactId>
       <version>${jitsi-desktop.version}</version>
       <scope>runtime</scope>
+    </dependency>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>jitsi-webrtcvadwrapper</artifactId>
+      <version>1.0-20200414.144800-2</version>
     </dependency>
     <dependency>
       <groupId>${project.groupId}</groupId>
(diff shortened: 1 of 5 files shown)


src/main/java/org/jitsi/jigasi/transcription/SilenceFilter.java
```diff
@@ -0,0 +1,111 @@
+package org.jitsi.jigasi.transcription;
```
damencho, line 1: Can you add the headers, please? Thank you.

pom.xml
```diff
@@ -155,6 +155,11 @@
       <artifactId>jitsi-util</artifactId>
       <version>${jitsi-desktop.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.jitsi</groupId>
+      <artifactId>jitsi-webrtcvadwrapper</artifactId>
+      <version>1.0-SNAPSHOT</version>
```
bgrozev, line 161: Do you want `1.0-20200414.144800-2` here?

bgrozev: Oh, that's duplicate

paweldomas: oh I will remove it

paweldomas: something went wrong on cherry pick/rebase, thanks for noticing

paweldomas: https://github.com/jitsi/jigasi/pull/263

src/main/java/org/jitsi/jigasi/transcription/Participant.java
```diff
@@ -585,21 +614,49 @@ public boolean isCompleted()
...
+                   toBuffer = audio;
+               }
+
+               try
+               {
+                   buffer.put(toBuffer);
+               }
+               catch (BufferOverflowException | ReadOnlyBufferException e)
+               {
+                   sendRequest(audio);
```
bgrozev, line 650: Should we:
a) send this *after* sending the rest of the contents of `buffer` (line 656)?
b) use `toBuffer` instead of `audio` in case we passed through the `getSpeechWindow` branch (line 623)?

nikvaessen: If we catch these Exceptions I think we should:
1) `sendRequest(buffer.array())`
2) `buffer.clear()`
3) `sendRequest(toBuffer)`

nikvaessen: Actually, maybe the audio in the buffer is stale....

nikvaessen: Maybe we should clear the buffer at line 627?

nikvaessen: Actually the buffer is cleared when there is a new speech window (line 631) so stale audio will never be sent. So the answer stays the same.

PR #265, opened 2020-04-15
feat(transcription): improve handling of buffer

nikvaessen:
Per the comments of Boris in #261 the `BufferOverflowException` was not handled properly.

The buffer is also cleared whenever a "silent" window is detected. I think it's equivalent to clearing the buffer whenever a new "speech" window is detected expect for the fact that there is not stale audio in memory, so it's a bit more efficient.

### DIFF
--- a/src/main/java/org/jitsi/jigasi/transcription/Participant.java
+++ b/src/main/java/org/jitsi/jigasi/transcription/Participant.java
@@ -624,11 +624,11 @@ public class Participant
                    silenceFilter.giveSegment(audio);
                    if(silenceFilter.shouldFilter())
                    {
+                       buffer.clear();
                        return;
                    }
                    else if(silenceFilter.newSpeech())
                    {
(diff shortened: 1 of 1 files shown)


paweldomas:
eh I don't know when I will manage to try it out... seems my jigasi got crazy spitting `2020-04-16 13:33:37.958 WARNING: [144] org.jitsi.jigasi.transcription.Transcriber.log() dsfdsfdsfdsfvfdsf@conference.pawel2.jitsi.net: reading from SSRC 997520269 while it is not known as a participant
2020-04-16 13:33:37.958 WARNING: [144] org.jitsi.jigasi.transcription.Transcriber.log() dsfdsfdsfdsfvfdsf@conference.pawel2.jitsi.net: reading from SSRC 3813343611 while it is not known as a participant` all over the place and I can't figure it out. It doesn't seem to be directly related to the PR.

PR #273, opened 2020-04-24
Add call direction header

### LINKED ISSUE #272: Call direction
In order to react to the call direction on xmpp world it would be great to get jigasi passing the information along the presence stanza.

For information, our use case is setting different max occupant limits based on the call direction (like, max 5 slots for incoming calls)

jbguerraz:
This adds a header that inform about the call direction and so get it available on xmpp leg.

Close #272

### DIFF
--- a/src/main/java/org/jitsi/jigasi/SipGateway.java
+++ b/src/main/java/org/jitsi/jigasi/SipGateway.java
@@ -36,6 +36,21 @@ public class SipGateway
     implements RegistrationStateChangeListener
 {
     /**
+     * Name of 'header' attribute that hold the call direction.
+     */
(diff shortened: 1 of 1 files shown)


damencho:
Where is this header processed? On the sip side, outgoing and incoming calls are easy to distinguish ...

jbguerraz:
It's for usage on xmpp world.
Typically on a prosody module (see related issue)

jbguerraz:
at first (on v1.1), we added that info as an header extension directly on presence stanza from `joinConferenceRoom` method, but since you added this extra headers bag it looked more elegant that way :)

damencho:

PR #278, opened 2020-04-28
Make joinConference retryable

### LINKED ISSUE #277: Retry to join conference when it fails
If join conference fail, retry instead of hangup would be great.
In case of max occupant limit is reached, it'll allow to join the conference as soon as a slot become free, or if one has disabled the max occupant notification, then to end up on voicemail (instead of call ending with "busy" case). Also in case of temporary failure, it'll give a chance to the call to succeed.

FYI, our use case is : call a conference and if there is no slot free, leave a message on voicemail.

jbguerraz:
Closes #277

Default is to retry for 25 seconds. Most calls will endup on voicemail (if there is one) after 20 seconds.

It mostly adds a retry block around the try/catch of the `joinConference` method.
In `catch` case, it cancels the `inviteTimeout` (we added a set of thread to null in case of `cancel` as well to allow re-scheduling).
Finally it removes the hangup next to max occupants notification play (replaced the latch by a sleep to wait for end of play). The max occupant sound will be played in loop for 25 seconds with a 5 seconds interval (so twice) until there is a conference slot free or until the caller hangup.

### DIFF
--- a/src/main/java/org/jitsi/jigasi/JvbConference.java
+++ b/src/main/java/org/jitsi/jigasi/JvbConference.java
@@ -138,6 +138,16 @@ public class JvbConference
         = "org.jitsi.jigasi.NOTIFY_MAX_OCCUPANTS";
 
     /**
+     * The seconds to wait before retrying a join conference after a failure.
+     */
+    private static final long JOIN_CONFERENCE_RETRY_INTERVAL = 500;
+
+    /**
+     * The join conference retries limit.
+     */
+    private static final int JOIN_CONFERENCE_RETRY_LIMIT = 50;
+
+    /**
      * The default bridge id to use.
      */
     public static final String DEFAULT_BRIDGE_ID = "jitsi";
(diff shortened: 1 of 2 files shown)


damencho:
What about handling this complexity on the sip side, pushing a hangup reason and retry there ... What worries me there here that there are many problem scenarios with clearing calls and hanging up and adding one more complexity in the code worries me ...

jbguerraz:
I understand the point about Jigasi scope and the unix philosophy basically.
It could (should?) certainly join conference or hangup the call with an explicit reason.
Then announces could also be played by the, let's say, ipbx.
In our case we didn't (yet) deploy asterisk (we'll in the next 6 months), and so have Jigasi direct to the sip trunk.
OTOH, this is mostly nothing more than a loop around the `joinConference` code.
We also could restrict that retry to "max occupants" case only (actually, service-unavailable) if you prefer. Or we could forget that (we'll be sad only for few months after all, and maybe a bit more when we'll face the asterisk dialplan :D).
Up to you :)

jbguerraz:
@damencho what's your point of view ? is that retry loop really a blocker ? thank you a lot!

damencho:
Yeah, I don't like to add more complexity there ... it is enough complex to add retries, sorry.

jbguerraz:
@damencho ok. so be it. we'll maintain a fork then until we can add an ipbx (having callers get busy line signal isn't cool. they gonna contact support to ask what's wrong with the conference bridge).

PR #285, opened 2020-05-21
fix: Changes CallManager to detect unhealthy state.

damencho:
Uses a dynamic pool of threads and if max number is reached we mark
jigasi as unhealthy and fail calls.

### DIFF
--- a/src/main/java/org/jitsi/jigasi/CallManager.java
+++ b/src/main/java/org/jitsi/jigasi/CallManager.java
@@ -35,20 +35,28 @@ public class CallManager
 {
     private final static Logger logger = Logger.getLogger(CallManager.class);
 
-    private static final int POOL_SIZE = 5;
+    private static final int POOL_MAX_SIZE = 20;
 
     /**
      * The thread pool to serve all call operations like answer and hangup.
-     * We initialize it with the 5 threads and can grow up to 4 time that size.
+     * We initialize it with the 1 thread and can grow up to POOL_MAX_SIZE.
      */
-    private static ExecutorService threadPool
-        = new ThreadPoolExecutor(
-                POOL_SIZE, POOL_SIZE * 4,
-                60L, TimeUnit.SECONDS, // time to wait before clearing threads
-                new SynchronousQueue<>(),
-                new DaemonThreadFactory("Jigasi CallManager"));
+    private static ExecutorService threadPool = createNewThreadPool();
 
     private static boolean healthy = true;
+
+    /**
+     * Creates new thread pool.
+     * @return the newly created pool.
+     */
+    private static ExecutorService createNewThreadPool()
+    {
+        return new ThreadPoolExecutor(
+            1, POOL_MAX_SIZE,
+            60L, TimeUnit.SECONDS, // time to wait before clearing threads
+            new SynchronousQueue<>(),
+            new DaemonThreadFactory("Jigasi CallManager"));
+    }
 
     /**
      * Returns whether or not we consider this CallManager as healthy.
@@ -518,8 +526,9 @@ public class CallManager
         hangupCallThread.reasonCode = reasonCode;
         hangupCallThread.reason = reason;
 
-        // if we are unhealthy, let's process error hangups
-        if (!healthy && reasonCode > 400)
+        // if we are unhealthy, let's process the hangups with new threads
+        // so we can clean failed or ongoing calls
+        if (!healthy)
         {
             new Thread(hangupCallThread).start();
         }
@@ -688,7 +697,7 @@ public class CallManager
         if (!threadPool.isTerminated())
             throw new TimeoutException();
 
-        threadPool = Executors.newFixedThreadPool(POOL_SIZE);
+        threadPool = createNewThreadPool();
     }
 
     /**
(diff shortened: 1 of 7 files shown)


src/main/java/org/jitsi/jigasi/CallManager.java
```diff
@@ -25,21 +25,83 @@
...
     private static final int POOL_SIZE = 5;
 
+    /**
+     * The thread pool to serve all call operations like answer and hangup.
+     * We initialize it with the 5 threads and can grow up to 4 time that size.
+     */
     private static ExecutorService threadPool
-        = Executors.newFixedThreadPool(POOL_SIZE);
+        = new ThreadPoolExecutor(
+                POOL_SIZE, POOL_SIZE * 4,
```
bgrozev, line 46: Why do we need minimum 5 threads? They don't do a lot of work, and 1 should suffice.

src/main/java/org/jitsi/jigasi/CallManager.java
```diff
@@ -459,7 +518,15 @@ public synchronized static void hangupCall(Call   call,
         hangupCallThread.reasonCode = reasonCode;
         hangupCallThread.reason = reason;
 
-        threadPool.submit(hangupCallThread);
+        // if we are unhealthy, let's process error hangups
+        if (!healthy && reasonCode > 400)
```
bgrozev, line 522: Why the `reasonCode > 400` check? The only case where `healthy==false`, we suspect `submit()` will fail with an exception.

src/main/java/org/jitsi/jigasi/CallManager.java
damencho: Ok, I will update it.

src/main/java/org/jitsi/jigasi/CallManager.java
damencho: Ok, I will drop that check.

src/main/java/org/jitsi/jigasi/CallManager.java
```diff
@@ -623,4 +690,60 @@ public static synchronized void restartPool()
 
         threadPool = Executors.newFixedThreadPool(POOL_SIZE);
     }
+
+    /**
+     * A thread factory that creates daemon threads.
+     *
+     * @author Pawel Domas
+     */
+    private static class DaemonThreadFactory
```
bgrozev, line 699: Is this a copy/paste from jicofo...?

bgrozev: Please use `CustomizableThreadFactory` from jitsi-utils

src/main/java/org/jitsi/jigasi/CallManager.java
```diff
@@ -623,4 +690,60 @@ public static synchronized void restartPool()
...
+                if (threadName != null)
+                {
+                    t.setName(threadName + "-" + t.getId());
+                }
+            }
+            return t;
+        }
+    }
+
+    private static class DummyCallThread
```
bgrozev, line 729: This is not used anywhere?

src/main/java/org/jitsi/jigasi/SipGatewaySession.java
```diff
@@ -436,7 +445,16 @@ Exception onConferenceCallStarted(Call jvbConferenceCall)
...
         if (destination == null)
         {
-            CallManager.acceptCall(sipCall);
+            try
+            {
+                CallManager.acceptCall(sipCall);
+            }
+            catch(OperationFailedException e)
+            {
+                hangUpSipCall(e.getErrorCode(), "Cannot answer call");
```
bgrozev, line 454: Do we want to include the meessage as well (`e.getMessage()`)?

src/main/java/org/jitsi/jigasi/health/Health.java
```diff
@@ -110,6 +110,12 @@ public static void stop()
     public static void check()
         throws Exception
     {
+        if (!CallManager.isHealthy())
+        {
+            throw new Exception(
+                "CallManager blocked to answer or hangup calls.");
```
bgrozev, line 116: I think just `CallManager is not healthy` should suffice as an explanation. `Health.java` shouldn't care about the specifics

src/main/java/org/jitsi/jigasi/CallManager.java
damencho: Yes.

src/main/java/org/jitsi/jigasi/CallManager.java
damencho: Yes. Just give a name of the threads for the stack traces.

src/main/java/org/jitsi/jigasi/SipGatewaySession.java
damencho: Well, that ends up in the sip message ... I was thinking it maybe too long ...

src/main/java/org/jitsi/jigasi/CallManager.java
damencho: Nice, thanks.

src/main/java/org/jitsi/jigasi/CallManager.java
damencho: Oh, this is something I forgot to delete :) I was filling pools with this one while testing :P

damencho: It is deleted now.

damencho:
Thanks.

PR #292, opened 2020-06-08
Reports full connference id to callstats.

damencho:


src/main/java/org/jitsi/jigasi/JvbConference.java
```diff
@@ -369,14 +369,16 @@ private Localpart getResourceIdentifier()
                 }
                 catch (XmppStringprepException e)
                 {
-                    logger.error("The SIP URI is invalid to use an XMPP"
+                    logger.error(this.callContext
+                        + "The SIP URI is invalid to use an XMPP"
```
paweldomas, line 373: maybe add a space(both places)?

src/main/java/org/jitsi/jigasi/stats/StatsHandler.java
```diff
@@ -316,6 +311,24 @@ private void createAndStartCallPeriodicRunnable(
         int interval = accountID.getAccountPropertyInt(
             CS_ACC_PROP_STATISTICS_INTERVAL, DEFAULT_STAT_INTERVAL);
 
+        String conferenceName = callContext.getRoomName();
+        if (conferenceName.contains("@"))
+        {
+            conferenceName = Util.extractCallIdFromResource(conferenceName);
```
paweldomas, line 317: maybe it would be worth moving this logic to a `getConferenceName` getter inside the `CallContext`?

PR #302, opened 2020-07-07
Update quick-install URL

black-puppydog:
Was pointing to a document that just said "has moved". 

This way everyone saves a *whole* click! :)


damencho:
This should point to https://jitsi.org/qi actually.

PR #304, opened 2020-07-28
Added config option for google cloud langauge

mayank10j:
This commit adds support for changing default transcription language code using config variable
org.jitsi.jigasi.transcription.LANG_CODE=en-US inside file sip-communicator.properties


damencho:
This breaks the logic, so you request a speech to text for English:
https://github.com/jitsi/jigasi/blob/6df3db2c63ccae1ac89d0591e6fe9005603c8a62/src/main/java/org/jitsi/jigasi/transcription/Participant.java#L678
https://github.com/jitsi/jigasi/blob/6df3db2c63ccae1ac89d0591e6fe9005603c8a62/src/main/java/org/jitsi/jigasi/transcription/Participant.java#L142
and then will mark the result as a different language.   

What do you want to achieve?

PR #325, opened 2020-11-13
Drops component and configures mucs by default

damencho:


debian/postinst
```diff
@@ -59,13 +59,12 @@ case "$1" in
             JIGASI_SIPPWD=$(echo -n $RET | base64 -)
 
             # 8-chars random secret, alternative to pwgen 8
-            JIGASI_SECRET=`head -c 8 /dev/urandom | tr '\0-\377' 'a-zA-Z0-9a-zA-Z0-9a-zA-Z0-9a-zA-Z0-9@@@@####'`
+            JIGASI_XMPP_SECRET=`head -c 8 /dev/urandom | tr '\0-\377' 'a-zA-Z0-9a-zA-Z0-9a-zA-Z0-9a-zA-Z0-9@@@@####'`
```
bgrozev, line 62: In other projects we use:
```
function generateRandomPassword() {
    cat /dev/urandom | tr -dc 'a-zA-Z0-9' | head -c 16
}
```

I can do that in a follow up PR if you prefer not to do it here.

debian/postinst
```diff
@@ -142,23 +141,50 @@ case "$1" in
...
         PROSODY_HOST_CONFIG="/etc/prosody/conf.avail/$JVB_HOSTNAME.cfg.lua"
         if [ -f $PROSODY_HOST_CONFIG ]; then
-            JIGASI_PROSODY="Component \"callcontrol.$JVB_HOSTNAME\" component_secret = \"$JIGASI_SECRET\""
-            # update prosody config
-            if [ ! "$(grep "Component \"callcontrol.$JVB_HOSTNAME\"" $PROSODY_HOST_CONFIG)" ]; then
-                echo $JIGASI_PROSODY >> $PROSODY_HOST_CONFIG
-                invoke-rc.d prosody restart
+            USER_EXISTS_CHECK=`prosodyctl adduser jigasi@auth.$JVB_HOSTNAME < /dev/null || true`
+            if [ ! "$USER_EXISTS_CHECK" = "That user already exists" ]; then
+                prosodyctl register jigasi auth.$JVB_HOSTNAME $JIGASI_XMPP_SECRET || true
```
bgrozev, line 159: This is a bit confusing. The goal is to create a new user if one doesn't exists, and otherwise do what? Keep the previous password or change it? A log message might be useful.

I looked up the difference between `adduser` and `register`, and they should accomplish the same, except `register` is maintained for compatibility with `ejabberdctl`. Can we just pass the password to `adduser` and check the return status?

debian/postinst
```diff
@@ -142,23 +141,50 @@ case "$1" in
...
+            fi
+        fi
+
+        JICOFO_CONFIG="/etc/jitsi/jicofo/jicofo.conf"
+        if [ -f $JICOFO_CONFIG ]; then
+            # add brewery room to jicofo config if it is not there
+            RES=0
+            hocon -f $JICOFO_CONFIG get jicofo.jigasi.brewery > /dev/null 2>&1 || RES=$? || true
+            if [ $RES -eq 1 ]; then
+                hocon -f $JICOFO_CONFIG set jicofo.jigasi.brewery \"JigasiBrewery@internal.auth.meet.damencho.net\" || true
```
bgrozev, line 169: Oops, I think this needs the propert domain name

debian/postinst
```diff
@@ -142,23 +141,50 @@ case "$1" in
...
+            if [ ! "$USER_EXISTS_CHECK" = "That user already exists" ]; then
+                prosodyctl register jigasi auth.$JVB_HOSTNAME $JIGASI_XMPP_SECRET || true
+            fi
+        fi
+
+        JICOFO_CONFIG="/etc/jitsi/jicofo/jicofo.conf"
+        if [ -f $JICOFO_CONFIG ]; then
+            # add brewery room to jicofo config if it is not there
+            RES=0
+            hocon -f $JICOFO_CONFIG get jicofo.jigasi.brewery > /dev/null 2>&1 || RES=$? || true
```
bgrozev, line 167: Is this equivalent?
```
if ! hocon -f $JICOFO_CONFIG get jicofo.jigasi.brewery > /dev/null 2>&1 ;then
    hocon -f $JICOFO_CONFIG set jicofo.jigasi.brewery \"JigasiBrewery@internal.auth.meet.damencho.net\"
fi
```

jigasi-home/sip-communicator.properties
```diff
@@ -66,6 +66,75 @@ net.java.sip.communicator.impl.protocol.sip.acc1403273890647.DOMAIN_BASE=<<DOMAI
...
+net.java.sip.communicator.impl.protocol.jabber.acc-xmpp-1=acc-xmpp-1
+net.java.sip.communicator.impl.protocol.jabber.acc-xmpp-1.ACCOUNT_UID=Jabber:jigasi@auth.<<DOMAIN_BASE>>
+net.java.sip.communicator.impl.protocol.jabber.acc-xmpp-1.USER_ID=jigasi@auth.<<DOMAIN_BASE>>
+net.java.sip.communicator.impl.protocol.jabber.acc-xmpp-1.IS_SERVER_OVERRIDDEN=true
+net.java.sip.communicator.impl.protocol.jabber.acc-xmpp-1.SERVER_ADDRESS=127.0.0.1
+net.java.sip.communicator.impl.protocol.jabber.acc-xmpp-1.SERVER_PORT=5222
+#net.java.sip.communicator.impl.protocol.jabber.acc-xmpp-1.BOSH_URL=https://127.0.0.1/http-bind
+net.java.sip.communicator.impl.protocol.jabber.acc-xmpp-1.ALLOW_NON_SECURE=true
+#base64 AES keyLength:256 or 128
+net.java.sip.communicator.impl.protocol.jabber.acc-xmpp-1.PASSWORD=<<JIGASI_XMPPPWD>>
```
bgrozev, line 80: Can we name this more clearly? JIGASI_XMPP_PASSWORD_BASE64?

debian/postinst
```diff
@@ -142,23 +141,50 @@ case "$1" in
...
+                # now let's restart jicofo
+                systemctl restart jicofo
             fi
         fi
 
+        # Prints a warning when we think that brewery is not configured
+        if ! grep -q ".BREWERY" "$JIGASI_CONFIG" ;then
+            echo "------------------------------------------------"
+            echo ""
+            echo "MUC brewery not configured!"
```
bgrozev, line 180: Do we also log a clear warning when jigasi runs without a brewery configured?

debian/postinst
damencho: Well, register has a param with a password, adduser reads it from stdin. 
Well we create the user once and put the password in the template once, so all of the rest configures on the update just skip.

debian/postinst
damencho: shit :) will force push it

debian/postinst
damencho: Nope that doesn't work, I tested that first but it gives an error and the script is stopped

jigasi-home/sip-communicator.properties
damencho: 👍

debian/postinst
damencho: 👍

debian/postinst
damencho: Nope, I will try to add one.

debian/postinst
bgrozev: So the first time this runs the user does not exist. We add a user via `adduser`, and then also run `register` which updates the password?

And the subsequent runs when the user exists, we do nothing?

If so, I think we can simplify to just `adduser` with the password, e.g.:
```
echo $JIGASI_XMPP_SECRET | prosodyctl adduser jigasi@auth.$JVB_HOSTNAME || true
```

damencho: > And the subsequent runs when the user exists, we do nothing?

Yes.

> If so, I think we can simplify to just adduser with the password, e.g.:

But this will update the password on every update and will break it after the first one.

This code was borrowed from: https://github.com/jitsi/jitsi-meet/blob/29fa4c935e9449f000d23801e001663ab882ec74/debian/jitsi-meet-prosody.postinst#L131

debian/postinst
bgrozev: I don't understand why. This runs as expected for me:
```bash
$ cat test.sh
#!/bin/bash
set -e

echo "" > test.conf

if ! hocon -f test.conf get nonexistent.key > /dev/null 2>&1 ; then
	echo "The key is missing"
fi

echo "This still prints"
```

```
$ ./test.sh
The key is missing
This still prints
```

debian/postinst
bgrozev: So `adduser` with empty input fails with `That user already exists`, but `adduser` with non-empty input updates the password?

debian/postinst
damencho: Cause it is not bash that runs it. Ok let me change it in a commit so I will build the debian package and test it and will show you.

damencho: Ok, it works. I'm testing this the 5th time already not sure what was the problem before :(
but the post inst script was failing through the dpkg and was passing when I was executing it in bash console

jigasi-home/sip-communicator.properties
bgrozev: Sorry, one more instance of the old name left in the readme

bgrozev (approved):

bgrozev (dismissed):

bgrozev:
@damencho I'm going to change the brewery config from `jicofo.jigasi.brewery` to `jicofo.jigasi.brewery-jid`. Can you update it here please?

damencho:
Done, I haven't tested it though, but I suppose the - in the name will not bring any problems.

damencho:
@bgrozev Can you approve this so we can merge it as you had already merged the jicofo part of it :)

bgrozev (approved):

PR #326, opened 2020-11-24
Pass token to incoming call

kaveman-:
We needed tokens to be passed to incoming calls and the attached PR is a quick hack to add the feature. A review is appreciated. Thanks.

### DIFF
--- a/src/main/java/org/jitsi/jigasi/CallContext.java
+++ b/src/main/java/org/jitsi/jigasi/CallContext.java
@@ -82,13 +82,20 @@ public class CallContext
     private String roomPassword;
 
     /**
+     * Optional token (JWT) required to enter MUC room
+     * when using token based authentication.
+     */
+    private String authToken;
+
+    /**
      * Optional bosh url that we use to join a room with the
      * xmpp account.
      * The bosh URL is a pattern:
      * https://{host}{subdomain}/http-bind?room={roomName}
      * Call context take care of the parameters {host} and {subdomain}
      * replacing them with domain and if available subdomain separating them
-     * with '/', otherwise replaces subdomain with ''.
+     * with '/', otherwise replaces subdomain with ''. If authToken is
+     * set, then an additional parameter &token={authToken} is appended.
      */
     private String boshURL;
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/jitsi/jigasi/CallContext.java
```diff
@@ -416,6 +440,15 @@ private void update()
             // target room name
             boshURL = boshURL.replace("{subdomain}", subdomain);
         }
+
+        if (this.authToken != null)
+        {
+            String tokenParam = "&token=" + this.authToken;
+            if (!boshURL.endsWith(tokenParam))
```
damencho, line 447: Can you change it to just contains &token= to minimize string comparison.

damencho (changes requested):

src/main/java/org/jitsi/jigasi/CallContext.java
kaveman-: > Hi, thanks for your contribution!
> If you haven't already done so, could you please make sure you sign our CLA (https://jitsi.org/icla for individuals and https://jitsi.org/ccla for corporations)? We would unfortunately be unable to merge your patch unless we have that piece :(.

under process. will be done shortly.

sebastienleon:
Hi Damencho,
Please, find the CLA attached to this message.
Best regards
Sebastien
[Please_DocuSign_8x8_Corporate_CLA.pdf](https://github.com/jitsi/jigasi/files/5591195/Please_DocuSign_8x8_Corporate_CLA.pdf)

src/main/java/org/jitsi/jigasi/CallContext.java
kaveman-: > Can you change it to just contains &token= to minimize string comparison.

wouldn't it break if the BOSH_URL_PATTERN itself contains a "&token=" ? Also, how expensive is the endsWith() comparison? JWT tends to be long, yes, but how often are these comparisons made? update() only after the token is set and the token is set last so basically only once. no?

damencho: If token is set, don't set it twice, so that sounds ok. Adding two tokens you risk exceeding url size limits as there can be such on some services ...

damencho: > JWT tends to be long, yes, but how often are these comparisons made? 

Not much, but why doing it, if there is a quicker check.

kaveman-: > > JWT tends to be long, yes, but how often are these comparisons made?
> 
> Not much, but why doing it, if there is a quicker check.

done.

sebastienleon: Hi Damencho,
Do you need anything else to merge this ?
(we modified code according your request and uploaded the CLA)
(we need to package jigasi now and it will be easier if it is from your repos than from our fork) 
Best regards
Sebastien

damencho (approved):

PR #334, opened 2021-01-07
Updates signaling of features and presences in the room

damencho:


src/main/java/org/jitsi/jigasi/JvbConference.java
```diff
@@ -1846,6 +1856,19 @@ public void onPasswordReceived(String pwd)
         joinConferenceRoom();
     }
 
+    /**
+     * Creates a feature xmpp extension, that can be added to features and used in presence.
+     * @param var the value to be added.
+     * @return the extension element.
+     */
+    private static ExtensionElement getFeature(String var)
```
bgrozev, line 1890: nit: Statics on top, maybe name `createFeature`

src/main/java/org/jitsi/jigasi/JvbConference.java
```diff
@@ -1201,20 +1212,9 @@ private void processChatRoomMemberLeft(ChatRoomMember member)
...
-                        DiscoverInfo i = ServiceDiscoveryManager.getInstanceFor(getConnection())
-                            .discoverInfo(JidCreate.entityFullFrom(jidString));
-
-                        if (!i.containsFeature(SIP_GATEWAY_FEATURE_NAME))
-                        {
-                            return true;
-                        }
-                    }
-                    catch(Exception e)
+                    if (!jigasiChatRoomMembers.contains(m.getName()))
```
bgrozev: `return !jigasiChatRoomMembers.contains(m.getName())`?

I find all this hard to read, I'd do something like 
```
onlyJigasisInRoom = allMatch(m ->
  m.getName().equals(getResourceIdentifier().toString() 
  || m.getName().equals(gatewaySession.getFocusResourceAddr()) 
  || jigasiChatRoomMembers.contains(m.getName()));
 ```

Comment (bgrozev) [reply src/main/java/org/jitsi/jigasi/JvbConference.java]: There's also `ChatRoomMemberJabberImpl#getLastPresence()` that you can use to make a `isJigasi` utility an avoid saving jigasi names in a list.

Comment (damencho) [reply src/main/java/org/jitsi/jigasi/JvbConference.java]: There was something about it that prevents me from using it ... I need to double check

File: src/main/java/org/jitsi/jigasi/JvbConference.java [hunk 3]
```diff
@@ -1190,62 +1231,48 @@ private void processChatRoomMemberLeft(ChatRoomMember member)
             return;
         }
 
-        try
+        // if lobby is not enabled or single moderator mode is detected
+        // there is nothing to process
+        // but otherwise we will check whether there are
+        if (this.lobbyEnabled && !this.singleModeratorEnabled)
```
bgrozev, line 1237: The doc isn't finished. ill check whether there are ...?

src/main/java/org/jitsi/jigasi/JvbConference.java
damencho: I cannot use getLastPresence(), cause jitsi xmpp PP does not store the first presence ... only the updates after creating the member. I will leave it like this, as it needs more refactoring in jitsi-desktop.

bgrozev (approved):

PR #348, opened 2021-04-01
feat: Checks the jvb call for activity and drop on no media.

damencho:
In case of no activity we update the stats and drop the xmpp and sip call.

### DIFF
--- a/src/main/java/org/jitsi/jigasi/JvbConference.java
+++ b/src/main/java/org/jitsi/jigasi/JvbConference.java
@@ -24,6 +24,7 @@ import net.java.sip.communicator.service.protocol.jabber.*;
 import net.java.sip.communicator.service.protocol.media.*;
 import net.java.sip.communicator.util.Logger;
 import net.java.sip.communicator.util.*;
+import org.jitsi.impl.neomedia.*;
 import org.jitsi.jigasi.lobby.Lobby;
 import org.jitsi.jigasi.stats.*;
 import org.jitsi.jigasi.util.*;
@@ -51,6 +52,7 @@ import org.jxmpp.stringprep.*;
 import org.osgi.framework.*;
 
 import java.beans.*;
+import java.io.*;
 import java.util.*;
 
 import static org.jivesoftware.smack.packet.XMPPError.Condition.*;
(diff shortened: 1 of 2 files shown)


src/main/java/org/jitsi/jigasi/JvbConference.java
```diff
@@ -2121,4 +2136,52 @@ public void processStanza(Stanza stanza)
...
+            CallPeer peer = jvbCall.getCallPeers().next();
+            if (peer instanceof MediaAwareCallPeer)
+            {
+                MediaAwareCallPeer peerMedia = (MediaAwareCallPeer) peer;
+
+                CallPeerMediaHandler mediaHandler = peerMedia.getMediaHandler();
+                if (mediaHandler != null)
+                {
+                    MediaStream stream = mediaHandler.getStream(MediaType.AUDIO);
+                    if (stream != null && stream instanceof AudioMediaStreamImpl)
```
bgrozev, line 2164: This should be equivalent to just `stream instanceof AudioMediaStreamImpl`

src/main/java/org/jitsi/jigasi/JvbConference.java
```diff
@@ -2121,4 +2136,52 @@ public void processStanza(Stanza stanza)
...
+                    {
+                        try
+                        {
+                            // if there is no activity on the audio channel this means there is a problem
+                            // establishing the media path with the bridge so we can just fail the call
+                            if (((AudioMediaStreamImpl) stream).getLastInputActivityTime() == 0)
```
bgrozev, line 2170: Looking at the impl, it returns `-1` when there are no input streams. Do you want to include `-1` here, or are there special semantics for `0` that I didn't find?

src/main/java/org/jitsi/jigasi/JvbConference.java
```diff
@@ -2121,4 +2136,52 @@ public void processStanza(Stanza stanza)
...
+                            {
+                                Statistics.incrementTotalCallsJvbNoMedia();
+                                logger.error(callContext + " No activity on JVB conference call will stop");
+
+                                stop();
+                            }
+                        }
+                        catch(IOException e)
+                        {
+                            logger.error("", e);
```
bgrozev, line 2180: Add a message?

src/main/java/org/jitsi/jigasi/JvbConference.java
```diff
@@ -2121,4 +2136,52 @@ public void processStanza(Stanza stanza)
...
+        @Override
+        public void run()
+        {
+            // if the call was stopped before we check ignore
+            if (!started)
+            {
+                return;
+            }
+
+            CallPeer peer = jvbCall.getCallPeers().next();
```
bgrozev, line 2155: Should the cases of `peer == null` or `mediaHandler == null` or `stream == null` also trigger stopping the conference (there is probably *some* problem)?

damencho: Probably ... this checks came from another code where we do it early in the code and retry ... will change that.

src/main/java/org/jitsi/jigasi/JvbConference.java
damencho: Ok, I can change it to `<= 0`. The 0 came from my testing ... I changed the code in a way that DTLS was not established and this was returning 0 ...

damencho:
I think, fixed all comments.

bgrozev (approved):

PR #349, opened 2021-04-02
feat: drop jigasi callers without web users

jerry2013:
I use jigasi in a special "mode" where after all web users have left, all jigasi calls need to be dropped automatically.

I'm proposing this change so that I don't have to build my own all the time but I'm not sure if it's acceptable by the jitsi community.

### DIFF
--- a/jigasi-home/sip-communicator.properties
+++ b/jigasi-home/sip-communicator.properties
@@ -164,6 +164,10 @@ org.jitsi.jigasi.xmpp.acc.IS_FILE_TRANSFER_DISABLED=true
 # org.jitsi.jigasi.xmpp.acc.PASS=SOME_PASS
 # org.jitsi.jigasi.xmpp.acc.ANONYMOUS_AUTH=false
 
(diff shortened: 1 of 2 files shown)


damencho:

damencho:
I'm wondering about some other name of the option? ... The default is a little bit confusing with the double negative ... disallow_only = false ... WDYT?

jerry2013:
Yeah, I wrecked my brain on a name that is clear, but could not come up with anything clear - I'm totally open to ideas.

Maybe JIGASI_ONLY_MEETING (default truthy)? or STOP_MEETING_WITHOUT_WEB (default falsy)?

damencho:
Yeah, maybe rename it to "org.jitsi.jigasi.ALLOW_ONLY_JIGASIS_IN_ROOM" and that defaults to true.

jerry2013:
sorry for the delay - updated to ALLOW_ONLY_JIGASIS_IN_ROOM - and listed such in the config template

damencho (approved):

PR #360, opened 2021-05-20
fix: Handles participant left as sourceRemove.

damencho:
Also clears MediaStream Stats.

### DIFF
--- a/src/main/java/org/jitsi/jigasi/JvbConference.java
+++ b/src/main/java/org/jitsi/jigasi/JvbConference.java
@@ -1225,10 +1225,9 @@ public class JvbConference
             CallPeer peer;
             if (jvbCall != null && (peer = jvbCall.getCallPeers().next()) instanceof MediaAwareCallPeer)
             {
-                MediaAwareCallPeer peerMedia = (MediaAwareCallPeer) peer;
-                peerMedia.getConferenceMembers().stream().forEach(m ->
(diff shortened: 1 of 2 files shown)


src/main/java/org/jitsi/jigasi/JvbConference.java
```diff
@@ -1221,6 +1221,30 @@ else if(ChatRoomMemberPresenceChangeEvent.MEMBER_UPDATED
...
 
             jigasiChatRoomMembers.remove(member.getName());
+
+            CallPeer peer;
+            if (jvbCall != null && (peer = jvbCall.getCallPeers().next()) instanceof MediaAwareCallPeer)
+            {
+                MediaAwareCallPeer peerMedia = (MediaAwareCallPeer) peer;
+                peerMedia.getConferenceMembers().stream().forEach(m -> {
+
+                    String address = ((ConferenceMember)m).getAddress();
```
bgrozev, line 1231: Is the cast to `ConferenceMember` necessary?

damencho: Yep
```
                  /jigasi/src/main/java/org/jitsi/jigasi/JvbConference.java:[1231,39] cannot find symbol
[ERROR]   symbol:   method getAddress()
[ERROR]   location: variable m of type java.lang.Object
```

bgrozev (approved):

PR #387, opened 2021-11-14
Upgrade to Jitsi 2.14 (and everything that comes with it)

ibauersachs:


damencho (approved):

debian/control
```diff
@@ -3,13 +3,13 @@ Section: net
...
 Uploaders: Emil Ivov <emcho@jitsi.org>, Damian Minkov <damencho@jitsi.org>
-Build-Depends: debhelper, dh-systemd, java8-runtime-headless | java8-runtime | java11-runtime-headless | java11-runtime, maven
+Build-Depends: debhelper, dh-systemd, openjdk-11-jdk, maven
 Standards-Version: 3.9.3
 Homepage: https://jitsi.org
 
 Package: jigasi
 Architecture: any
-Depends: ${misc:Depends}, java8-runtime-headless | java8-runtime | java11-runtime-headless | java11-runtime, libxss1, openssl, ruby-hocon
+Depends: ${misc:Depends}, openjdk-11-jre-headless | openjdk-11-jre | java11-runtime-headless | java11-runtime, libxss1, openssl, ruby-hocon
```
ibauersachs, line 12: @damencho Why not use `default-jre-headless (>= 2:1.11)`? Everything >= Java 11 should work, including 17. No need to specifically request Java 11.

Also, why is the x11 screensaver needed? This seems wrong.

damencho: libxss1 was something from old times ... should I remove it? 

well for the java dependency I copied what we have for the bridge ... if I remember correctly the bridge has some GC garbage pauses with java17 and needs tuning (not that we explicitly set gc alg here as the bridge) ... that was my only constraint ...

damencho: and this combination was the only one that does not force java17. The order was important for ubuntu 22.04.

ibauersachs: I'd go for sanity now and revert to 11 only if there's specific issues. I guess the 8x8-hosted instances run on 11 anyway, so I doubt you'll see issues.

ibauersachs: > and this combination was the only one that does not force java17. The order was important for ubuntu 22.04.

https://packages.ubuntu.com/jammy/default-jre not sure where 17 comes from...?

damencho: It was getting it from java11-runtime (maybe) that was not available on 22.04 but was getting it from some suggestion some package has ... couldn't fully figure it out... 
Ok, I will switch it to default-jre-headless (>= 2:1.11) and will drop the libxss

damencho: https://github.com/jitsi/jigasi/pull/413

PR #388, opened 2021-11-19
fix: Fixes ssrc rewrite when creating outgoing calls.

damencho:
Without rewriting the ssrc jigasi pass through the one incoming from the sip provider and browsers doing ssrc validation will fail to play the audio (Safari, Firefox and chrome with unified plan) as we had announced another ssrc earlier when setting up the call.

### DIFF
--- a/src/main/java/org/jitsi/jigasi/SipGatewaySession.java
+++ b/src/main/java/org/jitsi/jigasi/SipGatewaySession.java
@@ -410,31 +410,29 @@ public class SipGatewaySession
         if (destination == null)
         {
             incomingCall.setConference(sipCall.getConference());
+        }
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/jitsi/jigasi/SipGatewaySession.java
```diff
@@ -917,6 +915,7 @@ private boolean addSsrcRewriter(CallPeer peer)
                 MediaStream stream = mediaHandler.getStream(MediaType.AUDIO);
                 if (stream != null)
                 {
+System.err.println("oooooooooooo " + stream.getLocalSourceID());
```
jallamsetty1, line 918: forgot to delete this ?

jallamsetty1 (approved):

PR #398, opened 2022-02-15
Fix transcriber problem when compiled on java11 and ran on java8

damencho:


bgrozev (approved):
Can you re-word the commit message when meriging (it reads as if it fixes a failure to compile)

PR #401, opened 2022-03-04
Tenant header option

damencho:
Adds an option for new sip header to pass not only the domain, but and the tenant.
Adds muc prefix to cover cases when docker is used and the muc component address is `muc.meet.jitsi`.
Make sure we use the muc jid to construct any jid we will be using, for joining, for querying jicofo or hosts for av moderation or lobby.

### DIFF
--- a/src/main/java/org/jitsi/jigasi/CallContext.java
+++ b/src/main/java/org/jitsi/jigasi/CallContext.java
@@ -17,7 +17,7 @@
  */
 package org.jitsi.jigasi;
 
-import org.jitsi.jigasi.util.*;
+import net.java.sip.communicator.util.*;
 import org.jitsi.utils.*;
 import org.jxmpp.jid.*;
 import org.jxmpp.jid.impl.*;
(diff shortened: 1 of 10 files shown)


src/main/java/org/jitsi/jigasi/CallContext.java
```diff
@@ -55,7 +69,16 @@
...
+    private EntityBareJid roomJid;
+
+    /**
+     * The domain extracted from the room jid, without the tenant part.
+     * e.g. if room jid is roomname@muc.tenant.domain.com or roomname@muc.domain.com then <tt>roomJidDomain</tt> will
+     * have the value of domain.com.
+     * In case of docker <tt>domain</tt> can be the domain name of the deployment (used to construct the bosh url),
+     * but the room jid can be roomname@muc.tenant.meet.jitsi or roomname@muc.meet.jitsi.
+     */
+    private String roomJidDomain;
```
gpolitis, line 81: _Maybe_ rename this to `roomJidTld` (tld stands for top-level domain).

src/main/java/org/jitsi/jigasi/CallContext.java
```diff
@@ -152,20 +175,36 @@ public CallContext(Object source)
 
     /**
      * The room name of the MUC room that holds JVB conference call.
-     * @return the room name.
+     * Can just room name or the jid (roomname@conference.tenant.domain.net).
+     *
+     * @return the room name or jid.
```
gpolitis, line 180: This can't be just the room name, right?

damencho: Yep.

gpolitis (approved):
LGTM

PR #413, opened 2022-05-17
feat: Sets jre11 as minimum required.

damencho:
Allows running it with java17. Drops unused dependency.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -9,7 +9,7 @@ Homepage: https://jitsi.org
 
 Package: jigasi
 Architecture: any
-Depends: ${misc:Depends}, openjdk-11-jre-headless | openjdk-11-jre | java11-runtime-headless | java11-runtime, libxss1, openssl, ruby-hocon
+Depends: ${misc:Depends}, default-jre-headless (>= 2:1.11), openssl, ruby-hocon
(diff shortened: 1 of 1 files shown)


debian/control
```diff
@@ -9,7 +9,7 @@ Homepage: https://jitsi.org
 
 Package: jigasi
 Architecture: any
-Depends: ${misc:Depends}, openjdk-11-jre-headless | openjdk-11-jre | java11-runtime-headless | java11-runtime, libxss1, openssl, ruby-hocon
+Depends: ${misc:Depends}, default-jre-headless (>= 2:1.11), openssl, ruby-hocon
```
ibauersachs, line 12: What's with the ruby-hocon thing?

damencho: https://github.com/jitsi/jigasi/blob/296b7cd94c0712d586737b1e59dd8d12f453a475/debian/postinst#L163

damencho: And of course the:
https://github.com/jitsi/jigasi/blob/296b7cd94c0712d586737b1e59dd8d12f453a475/debian/postinst#L164

ibauersachs: Urgh.

ibauersachs (approved):

PR #429, opened 2022-06-10
Configurable Google Speech to Text transcription

kratu92:
Options to enable punctuation and profanity filter to the results obtained when using Google Speech to Text for transcriptions.
Added option to enable the reception of interim results instead of only final results when using Google Speech to Text for transcriptions.
This options may be set in the sip-communicator.properties:
org.jitsi.jigasi.transcription.ENABLE_GOOGLE_AUTOMATIC_PUNCTUATION=false
org.jitsi.jigasi.transcription.ENABLE_GOOGLE_PROFANITY_FILTER=false
org.jitsi.jigasi.transcription.ENABLE_GOOGLE_INTERIM_RESULTS=false

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,8 +131,11 @@ By default, the properties
 `org.jitsi.jigasi.transcription.ENABLE_GOOGLE_AUTOMATIC_PUNCTUATION=false`
 and
 `org.jitsi.jigasi.transcription.ENABLE_GOOGLE_PROFANITY_FILTER=false`
+and
+`org.jitsi.jigasi.transcription.ENABLE_GOOGLE_INTERIM_RESULTS=false`
 in
 `jigasi-home/sip-communicator.properties`
+disable automatic punctuation, profanity filter and interim results for the transcription.
 To change this, simply set the desired property to `true` or `false`.
 
 Vosk configuration
--- a/README.md
+++ b/README.md
@@ -126,6 +126,12 @@ sudo apt-get update && sudo apt-get install google-cloud-sdk google-cloud-sdk-ap
 gcloud init
 gcloud auth application-default login
 ```
+It is possible to enable or disable the functionality of Google Cloud Speech to Text.
+By default, the properties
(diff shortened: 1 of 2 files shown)


nikvaessen:
If I remember correctly, without interum results, transcriptions can take 2 to three seconds to appear (end of sentence). If this is the case, what is the benefit of having a feature flag (`org.jitsi.jigasi.transcription.ENABLE_GOOGLE_INTERIM_RESULTS`) which decreases the UX? What is the use-case for setting `org.jitsi.jigasi.transcription.ENABLE_GOOGLE_INTERIM_RESULTS=False`

kratu92:
Setting org.jitsi.jigasi.transcription.ENABLE_GOOGLE_INTERIM_RESULTS=False is for the current default behaviour (before my change), when you don't receive the result until the sentence is finished,  so that is why I left it as the default value. 
Changing it to true will send the partial sentences while the active speaker is speaking (Vosk had this feature).
It can take a couple of seconds if the sentence is long, and we found out that people with hearing disability prefer to read the transcription while they see the person speaking than wait until the sentence has finished, even if the transcription changes a bit and is updated later. Also people with cognitive disability may find it confusing to see the person speaking and not see the transcription for a while.
This way you have the option to have the default functionality, before the change (leaving it false) or have the partial results (changing it to true).

bgrozev:
There's existing code to enable interim results, look for `RETRIEVE_INTERIM_RESULTS`.

Unless I'm missing something, your addition of `P_NAME_ENABLE_GOOGLE_INTERIM_RESULTS` does one thing only: when you enable the flag, it forces TranscriptionResult.isInterim to `false` regardless of whether the result is interim or not. My guess is the existing code fails to do something with interim results for you, but I wuld prefer to chase down the issue instead of forcing `isInterim=false`. Perhaps [this is where the problem stems from](https://github.com/jitsi/jigasi/blob/master/src/main/java/org/jitsi/jigasi/transcription/RemotePublisherTranscriptionHandler.java#L60)?

bgrozev:
This looks good to me, but I'll let @damencho review.

dnna:
Having automatic punctuation would be really nice. Did something block this PR?

PR #441, opened 2022-08-19
feat: enable cc translation

tamasdomokos:


src/main/java/org/jitsi/jigasi/TranscriptionGatewaySession.java
```diff
@@ -296,6 +296,8 @@ void notifyChatRoomMemberLeft(ChatRoomMember chatMember)
     void notifyChatRoomMemberUpdated(ChatRoomMember chatMember, Presence presence)
     {
         super.notifyChatRoomMemberUpdated(chatMember, presence);
+        String identifier = getParticipantIdentifier(chatMember);
+        this.transcriber.updateParticipant(identifier, chatMember);
```
damencho, line 300: Can you add a comment why is this done. ... I guess to update a language change coming in the presence ...

damencho (changes requested):
Just a small one.

damencho (approved):

PR #443, opened 2022-09-22
fix: source language equals target language no translation needed

tamasdomokos:
When the source is equal with the target language we don't need to translate

### DIFF
--- a/src/main/java/org/jitsi/jigasi/transcription/Transcriber.java
+++ b/src/main/java/org/jitsi/jigasi/transcription/Transcriber.java
@@ -343,7 +343,9 @@ public class Transcriber
                     String language
                         = translationLanguageExtension.getTranslationLanguage();
 
-                    this.updateParticipantTargetLanguage(identifier, language);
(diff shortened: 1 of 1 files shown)


src/main/java/org/jitsi/jigasi/transcription/Transcriber.java
```diff
@@ -343,7 +343,10 @@ public void updateParticipant(String identifier,
                     String language
                         = translationLanguageExtension.getTranslationLanguage();
 
-                    this.updateParticipantTargetLanguage(identifier, language);
+                    if (!participant.getSourceLanguage().equals(language))
```
damencho, line 346: What is getSourceLanguage() returns null? CAn you add and that check.

damencho (approved):

src/main/java/org/jitsi/jigasi/transcription/Transcriber.java
tamasdomokos: added the null check

PR #445, opened 2022-09-22
Maintains stable UUID for Vosk partial transcription results

TomDurrant-yt:


damencho:

src/main/java/org/jitsi/jigasi/transcription/VoskTranscriptionService.java
```diff
@@ -212,15 +215,16 @@ public void onMessage(String msg)
             {
                 partial = false;
                 result = obj.getString("text");
+                lastUUID = UUID.randomUUID();
             }
-            if (!result.isEmpty() && !result.equals(lastResult))
+            if (!result.isEmpty() && (!partial || !result.equals(lastResult)))
```
damencho, line 220: This means that partial results will not be sent, right?
I like the other approaches better as it is closer https://github.com/jitsi/jigasi/pull/449 to the other transcription service we have. Please comment there if you do not agree with something.
Thank you.

TomDurrant-yt: Partial results will still be sent. This is to ensure that final results are always sent, even if the text is the same, so that the client can know when the transcription is finished. I have added a comment requesting the same change be made in #449

damencho: 👍 Thank you.

damencho:
https://github.com/jitsi/jigasi/pull/449
Thanks for the PR review @TomDurrant-yt

PR #447, opened 2022-09-27
Adds a localhost:8080/debug endpoint to the REST API.

gpolitis:
The goal of this PR is to add a `localhost:8080/debug` endpoint to the REST API that (as a first step) lists the unique meeting ids of the conferences that a Jigasi instance has joined. This information will be used to discover Jigasis in rtcstats:

```json
{
  "gateways": {
    "1921781245": {
      "sessions": {
        "1512XXXXXXX (+1512XXXXXXX)": {
          "jvbConference": {
            "meetingUrl": "https://meet.example.org/confname",
            "meetingId": "776ef674-8fdb-42f1-8083-523cefbc3a96"
          }
        }
      }
    }
  }
}
```

### DIFF
--- a/src/main/java/org/jitsi/jigasi/JvbConference.java
+++ b/src/main/java/org/jitsi/jigasi/JvbConference.java
@@ -47,6 +47,7 @@ import org.jivesoftware.smackx.disco.packet.*;
 import org.jivesoftware.smackx.muc.packet.*;
 import org.jivesoftware.smackx.nick.packet.*;
 import org.jivesoftware.smackx.xdata.packet.*;
+import org.jivesoftware.smackx.xdata.*;
 import org.jxmpp.jid.*;
 import org.jxmpp.jid.impl.*;
 import org.jxmpp.jid.parts.*;
@@ -126,6 +127,11 @@ public class JvbConference
      */
     public static final String LOCAL_REGION_PNAME
         = "org.jitsi.jigasi.LOCAL_REGION";
+
+    /**
+     * The name of the (unique) meeting id field in the MUC data form.
+     */
+    private static final String DATA_FORM_MEETING_ID_FIELD_NAME = "meeting_id";
 
     /**
      * The milliseconds to wait before check the jvb side of the call for activity.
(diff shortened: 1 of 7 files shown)


src/main/java/org/jitsi/jigasi/SipGateway.java
```diff
@@ -188,9 +193,32 @@ public SipGatewaySession createOutgoingCall(CallContext ctx)
...
+        = Collections.synchronizedList(new ArrayList<WeakReference<SipGatewaySession>>());
+    
+    @Override
+    public OrderedJsonObject getDebugState()
+    {
+        OrderedJsonObject debugState = super.getDebugState();
+        weakSessions.forEach(ref -> {
+            SipGatewaySession session = ref.get();
+            if (session != null) {
+                debugState.put(session.hashCode(), session.getDebugState());
```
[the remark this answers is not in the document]
gpolitis: Thanks Boris, that's a good point, will do! Damencho also mentioned we don't need the weakSessions field, as we already have a sessions field inside the `AbstractGateway` so I'll shuffl things around tomorrow.

src/main/java/org/jitsi/jigasi/JvbConference.java
```diff
@@ -843,6 +854,9 @@ public void joinConferenceRoom()
                 updateFromRoomConfiguration();
             }
 
+            // set the unique meeting id from the muc configuration.
+            updateMeetingId();
```
damencho, line 858: Can we remove this from here and be getting that on the first time the meetingID is needed in a get method or something. 
The point is we will be slowing down the join conference by doing a disco-info and waiting for the reply Joining is already slow ... :)

src/main/java/org/jitsi/jigasi/JvbConference.java
```diff
@@ -1954,6 +2009,7 @@ public void processStanza(Stanza stanza)
             if (mucUser.getStatus().contains(MUCUser.Status.create(104)))
             {
                 updateFromRoomConfiguration();
+                updateMeetingId();
```
damencho, line 2012: Why do we need this, the id never changes. 
And this will do a second query for the room configuration ... another iq and response.

damencho (changes requested):

src/main/java/org/jitsi/jigasi/JvbConference.java
gpolitis: Good point, thanks damencho!

PR #451, opened 2022-10-05
fix: re-establish connection with transcription service if session is prematurely ended

AubynKen:
This PR fixes the issue mentioned in the thread below:
https://community.jitsi.org/t/jigasi-randomly-starts-sending-raw-vosk-json-response-as-subtitles/117831

The problem came from lost websocket connection between jigasi and the VOSK transcription service.
Before the change, a participant whose audio should be transcribed switches to the single request mode when the session is prematurely ended. The PR allows Jigasi to check whether streaming is supported by the transcription service, and if it is the case, re-establish the streaming session.

### DIFF
--- a/src/main/java/org/jitsi/jigasi/transcription/Participant.java
+++ b/src/main/java/org/jitsi/jigasi/transcription/Participant.java
@@ -660,9 +660,15 @@ public class Participant
             {
                 session.sendRequest(request);
             }
+            else if (transcriber.getTranscriptionService().supportsStreamRecognition())
+            // re-establish prematurely ended streaming session
+            {
+                session = transcriber.getTranscriptionService()
+                        .initStreamingSession(this);
+                session.addTranscriptionListener(this);
+            }
             else
             // fallback if TranscriptionService does not support streams
-            // or session got ended prematurely
             {
                 // FIXME: 22/07/17 This just assumes given BUFFER_LENGTH
                 // is long enough to get decent audio length. Also does


AubynKen:
I have signed the CLA.
This PR is part of a project by France Université Numérique to improve the accessibility of the French government's video conference platform.

charles-zablit:
Hello @damencho,

I'm bumping this PR as it has been stale for a while.
Is there anything we can do to improve this PR?

Thanks :)

src/main/java/org/jitsi/jigasi/transcription/Participant.java
```diff
@@ -660,9 +660,15 @@ private void sendRequest(byte[] audio)
             {
                 session.sendRequest(request);
             }
+            else if (transcriber.getTranscriptionService().supportsStreamRecognition())
+            // re-establish prematurely ended streaming session
+            {
+                session = transcriber.getTranscriptionService()
+                        .initStreamingSession(this);
+                session.addTranscriptionListener(this);
+            }
```
damencho, line 669: Shouldn't we here send the current request so we do not lose it?

charles-zablit: We have tried that, however, if the StreamingSession is not re-established immediately (host is restarting for instance), `sendRequest` will error out (`NullPointerException` )  and it will do that until the connexion is restablished. From our tests, this introduces undefined behaviour.

damencho: But the code you have added initializes the session again(re-establish), and once that is done calling sendRequest will not work. Is this what you mean?

charles-zablit: Exactly.
Because if the initialization fails, sendRequest will be called and the transcription service will not be listening, resulting in a `NullPointerException`

charles-zablit: The thing is, without this fix, the connection can just randomly break and it will try to use the `sendSingleRequest` mode which is not supported by Google or VOSK, it just produces gibberish results.

damencho: But what about when participant left and the session is ended? 
https://github.com/jitsi/jigasi/blob/78258390f4f35cbb562d3cd320b6bcf1188e412c/src/main/java/org/jitsi/jigasi/transcription/Participant.java#L539
This will re-establish the session ... which doesn't sound good

rpurdel:
@damencho  This totally makes sense, in my opinion at least. And it also doesn't impact my implementation. The only thing I'm wondering is if we should still keep the sendSingleRequest method altogether? I mean if the streaming fails for any of the transcription services, why should sending single requests to the same service work?

damencho:
@rpurdel I agree, if we merge this basically the single request code will never be reached. I will go and merge this and we can do a follow up PR to drop that, after merging the whisper impl.

charles-zablit:
> I mean if the streaming fails for any of the transcription services, why should sending single requests to the same service work?

That was our reasoning when we were building this with @AubynKen 👍

damencho:
Thanks @charles-zablit @AubynKen

PR #455, opened 2022-10-17
LibreTranslate live translation service

AubynKen:
This is my initial attempt to integrate LibreTranslate into jigasi.

Currently, there are two transcription services supported in Jigasi: Google Cloud transcription service and Vosk transcription service, the ladder being open-source. However, only Google translate is supported for live translation of transcript/ subtitles. 

This pull request integrates the open source offline translation service [LibreTranslate](https://github.com/LibreTranslate/LibreTranslate) into jigasi, allowing users hosting Jitsi to use an alternative translation service to Google's translation API.

The LibreTranslate server supports more than 20 languages and 56 language-to-language models. It can be run offline (once the models are downloaded), and has [actively maintained container images](https://hub.docker.com/r/libretranslate/libretranslate) on Docker Hub.

### DIFF
--- a/src/main/java/org/jitsi/jigasi/transcription/LibreTranslateTranslationService.java
+++ b/src/main/java/org/jitsi/jigasi/transcription/LibreTranslateTranslationService.java
@@ -1,0 +1,112 @@
+package org.jitsi.jigasi.transcription;
+
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+import org.apache.http.HttpResponse;
+import org.apache.http.client.methods.HttpPost;
+import org.apache.http.entity.ContentType;
+import org.apache.http.entity.StringEntity;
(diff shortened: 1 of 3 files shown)


src/main/java/org/jitsi/jigasi/transcription/LibreTranslateTranslationService.java
```diff
@@ -0,0 +1,112 @@
+package org.jitsi.jigasi.transcription;
```
damencho, line 1: Please add the license header like the rest of the files have it.

damencho (changes requested):
Also, I think your code is 80 columns wide, right? Can you adjust it to 120?
We try to keep it 120 for the new code.
https://github.com/jitsi/jigasi/blob/4964b522a18df7ac0296f3a580c2d24b11cc82bd/checkstyle.xml#L9
Thank you.

PR #457, opened 2022-10-19
feat: add VOSK routing

charles-zablit:
This PR implements language routing capabilities for the VOSK transcription service.

A VOSK model only allows one language, so each VOSK service can only transcribe one language. With language routing, we can assign a language to a websocket URL, which points to the correct VOSK container as shown in the image below:

Jigasi-VOSK

Please let me know if I can provide any additional information/modifications.

### DIFF
--- a/src/main/java/org/jitsi/jigasi/transcription/VoskTranscriptionService.java
+++ b/src/main/java/org/jitsi/jigasi/transcription/VoskTranscriptionService.java
@@ -23,8 +23,6 @@ import org.eclipse.jetty.websocket.client.*;
 import org.json.*;
 import org.jitsi.jigasi.*;
 import org.jitsi.utils.logging.*;
-import org.json.simple.parser.JSONParser;
-import org.json.simple.parser.ParseException;
 
 import javax.media.format.*;
 import java.io.*;
@@ -73,13 +71,15 @@ public class VoskTranscriptionService
     /**
      * Assigns the websocketUrl to use to websocketUrl by reading websocketUrlConfig;
      */
-    private void generateWebsocketUrl(Participant participant) throws ParseException {
+    private void generateWebsocketUrl(Participant participant)
+        throws org.json.simple.parser.ParseException
+    {
         if (!supportsLanguageRouting())
         {
             websocketUrl = websocketUrlConfig;
         }
 
-        JSONParser jsonParser = new JSONParser();
+        org.json.simple.parser.JSONParser jsonParser = new org.json.simple.parser.JSONParser();
         Object obj = jsonParser.parse(websocketUrlConfig);
         JSONObject languageMap = (JSONObject) obj;
         websocketUrl = languageMap.optString(participant.getSourceLanguage());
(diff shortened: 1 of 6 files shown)


src/main/java/org/jitsi/jigasi/transcription/VoskTranscriptionService.java
```diff
@@ -23,6 +23,8 @@
 import org.json.*;
 import org.jitsi.jigasi.*;
 import org.jitsi.utils.logging.*;
+import org.json.simple.parser.JSONParser;
+import org.json.simple.parser.ParseException;
```
damencho, line 27: Please use the star import.

src/main/java/org/jitsi/jigasi/transcription/VoskTranscriptionService.java
```diff
@@ -64,15 +66,32 @@
...
 
-    private final String websocketUrl;
+    private String websocketUrlConfig;
+
+    private String websocketUrl;
+
+    /**
+     * Assigns the websocketUrl to use to websocketUrl by reading websocketUrlConfig;
+     */
+    private void generateWebsocketUrl(Participant participant) throws ParseException {
```
damencho, line 76: This { will fail the code style check.

damencho (changes requested):
LGTM, will take a deeper look later. Just a few style issues I saw.

damencho:
And maybe some of the new code is column 80 compatible, not sure :) Make sure it is 120 :) Not a big deal though ...

src/main/java/org/jitsi/jigasi/transcription/VoskTranscriptionService.java
charles-zablit: For some reason it didn't, but it's fixed now 👍

src/main/java/org/jitsi/jigasi/transcription/VoskTranscriptionService.java
```diff
@@ -64,15 +64,34 @@
...
+    {
+        if (!supportsLanguageRouting())
+        {
+            websocketUrl = websocketUrlConfig;
+        }
+
+        org.json.simple.parser.JSONParser jsonParser = new org.json.simple.parser.JSONParser();
+        Object obj = jsonParser.parse(websocketUrlConfig);
+        JSONObject languageMap = (JSONObject) obj;
+        websocketUrl = languageMap.optString(participant.getSourceLanguage());
```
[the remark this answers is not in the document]
damencho: Does this code supports the old style of config?

`org.jitsi.jigasi.transcription.vosk.websocket_url=ws://localhost:2700`?

src/main/java/org/jitsi/jigasi/transcription/VoskTranscriptionService.java
damencho: Hum, then I need to revisit the config :)

src/main/java/org/jitsi/jigasi/transcription/VoskTranscriptionService.java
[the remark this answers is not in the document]
charles-zablit: > Do we default to some language if there is no match?

Hmm no, but I will set `"en"` as default, as most users would have their interface language set to English.

> Does this code support the old style of config?

Yes, it will detect that it's not JSON and use the string as the WebSocket URL directly.

damencho (approved):

PR #475, opened 2023-03-06
feat: Use time-based uuid for transcription id.

damencho:
This allows sorting of the resulted messages based on the id, which is created when that message/phrase started.

### DIFF
--- a/src/main/java/org/jitsi/jigasi/transcription/GoogleCloudTranscriptionService.java
+++ b/src/main/java/org/jitsi/jigasi/transcription/GoogleCloudTranscriptionService.java
@@ -355,7 +355,7 @@ public class GoogleCloudTranscriptionService
             resultConsumer.accept(
                     new TranscriptionResult(
                             null,
-                            UUID.randomUUID(),
+                            Generators.timeBasedReorderedGenerator().generate(),
                             false,
                             request.getLocale().toLanguageTag(),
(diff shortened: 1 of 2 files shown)


src/main/java/org/jitsi/jigasi/transcription/GoogleCloudTranscriptionService.java
```diff
@@ -894,7 +895,7 @@ void terminateCurrentSession()
             this.languageTag = languageTag;
             this.debugName = debugName;
 
-            messageID = UUID.randomUUID();
+            messageID = Generators.timeBasedGenerator().generate();
```
bgrozev: Can you make `messageID` final please?

bgrozev: There's another instance of `UUID.randomUUID()` on line 358. Maybe update at least for consistency?

bgrozev (changes requested):

src/main/java/org/jitsi/jigasi/transcription/GoogleCloudTranscriptionService.java
damencho: Done.

bgrozev (approved):

PR #500, opened 2023-09-25
feat: Offload xmpp/smack processing threads.

damencho:
Make sure we do not send and/or wait for responses in the smack threads.

### DIFF
--- a/src/main/java/org/jitsi/jigasi/JvbConference.java
+++ b/src/main/java/org/jitsi/jigasi/JvbConference.java
@@ -48,6 +48,7 @@ import org.jivesoftware.smackx.muc.packet.*;
 import org.jivesoftware.smackx.nick.packet.*;
 import org.jivesoftware.smackx.xdata.packet.*;
 import org.jivesoftware.smackx.xdata.*;
+import org.json.simple.*;
 import org.jxmpp.jid.*;
 import org.jxmpp.jid.impl.*;
 import org.jxmpp.jid.parts.*;
@@ -57,6 +58,7 @@ import org.osgi.framework.*;
 import java.beans.*;
 import java.io.*;
 import java.util.*;
+import java.util.concurrent.*;
 
 import static net.java.sip.communicator.service.protocol.event.LocalUserChatRoomPresenceChangeEvent.*;
 import static org.jivesoftware.smack.packet.StanzaError.Condition.*;
@@ -77,7 +79,8 @@ public class JvbConference
                ChatRoomMemberPresenceListener,
                LocalUserChatRoomPresenceListener,
                CallPeerConferenceListener,
-               PropertyChangeListener
+               PropertyChangeListener,
+               OperationSetJitsiMeetTools.JitsiMeetRequestListener
 {
     /**
      * The logger.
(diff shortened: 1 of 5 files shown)


src/main/java/org/jitsi/jigasi/JvbConference.java
```diff
@@ -150,6 +153,12 @@ public class JvbConference
      */
     private String meetingId;
 
+    /**
+     * A thread pool used to offload xmpp execution in a new thread to avoid blocking
+     * xmpp threads.
+     */
+    public static final ExecutorService xmppExecutorPool = Util.createNewThreadPool("xmpp-executor-pool");
```
bgrozev, line 160: Do we not have a generic "io" thread pool in jigasi?

src/main/java/org/jitsi/jigasi/JvbConference.java
```diff
@@ -1081,6 +1100,11 @@ public void serviceChanged(ServiceEvent serviceEvent)
 
     @Override
     public void memberPresenceChanged(ChatRoomMemberPresenceChangeEvent evt)
+    {
+        xmppExecutorPool.execute(() -> memberPresenceChangedInternal(evt));
```
bgrozev, line 1104: These may now execute concurrently or out of order. Do they need to go on a queue?

src/main/java/org/jitsi/jigasi/sounds/SoundNotificationManager.java
```diff
@@ -624,7 +624,7 @@ public void notifyChatRoomMemberJoined(ChatRoomMember member)
     public void notifyChatRoomMemberLeft(ChatRoomMember member)
     {
         // if this is the sip hanging up (stopping) skip playing
-        if (gatewaySession.getJvbConference().isStarted()
+        if (gatewaySession.getJvbConference() != null && gatewaySession.getJvbConference().isStarted()
```
bgrozev, line 627: AbstractGatewaySession.jvbConference can change to null, so this can still produce an NPE (granted, rarely)

src/main/java/org/jitsi/jigasi/JvbConference.java
```diff
@@ -603,7 +612,7 @@ private synchronized void setXmppProvider(
 
         if (xmppProvider.isRegistered())
         {
-            joinConferenceRoom();
+            xmppExecutorPool.execute(this::joinConferenceRoom);
```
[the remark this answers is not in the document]
damencho: joinConferenceRoom() itself has one big try catch, that should catch everything...

src/main/java/org/jitsi/jigasi/JvbConference.java
damencho: Nothing, this becomes one for the xmpp handling.

src/main/java/org/jitsi/jigasi/JvbConference.java
damencho: I don't think that matters ...

src/main/java/org/jitsi/jigasi/sounds/SoundNotificationManager.java
damencho: Updated.

bgrozev (approved):

PR #501, opened 2023-09-27
feat: Adds an option to randomize user-id used when client logins.

damencho:
Updates to latest desktop lib dependency to fix ALWAYS_TRUST_MODE_ENABLED.

### DIFF
--- a/src/main/java/org/jitsi/jigasi/JvbConference.java
+++ b/src/main/java/org/jitsi/jigasi/JvbConference.java
@@ -181,6 +181,11 @@ public class JvbConference
     );
 
     /**
+     * Used for randomizing usernames if needed.
+     */
+    private static final Random RANDOM = new Random();
+
+    /**
      * Adds the features supported by jigasi to a specific
      * <tt>OperationSetJitsiMeetTools</tt> instance.
      * @return Returns the 'features' extension element that can be added to presence.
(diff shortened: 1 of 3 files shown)


src/main/java/org/jitsi/jigasi/JvbConference.java
```diff
@@ -1649,6 +1654,27 @@ else if ("org.jitsi.jigasi.xmpp.acc.BOSH_URL_PATTERN".equals(overridenProp))
...
+            else if ("org.jitsi.jigasi.xmpp.acc.USER_ID".equals(overridenProp)
+                && JigasiBundleActivator.getConfigurationService().getBoolean(overridePrefix + ".UNIQUE_USER_ID", false))
+            {
+                try
+                {
+                    Jid jid = JidCreate.from(value);
+                    long random = RANDOM.nextInt();
+                    Jid newJid = JidCreate.entityBareFrom(
+                            Localpart.from(
+                                    jid.getLocalpartOrNull() + "-" + String.format("%8h", random).replace(' ', '0')),
```
bgrozev, line 1666: Why do you need `.replace(' ', '0')`? Why "%h" (it calls hashCode) instead of %x?

damencho: I used your code from here: https://github.com/jitsi/jigasi/blob/fb76d173eb643b94fb7ee10721af1a1647127ee9/src/main/java/org/jitsi/jigasi/CallContext.java#L460

bgrozev (approved):

PR #506, opened 2023-10-19
Timestamp improvements

megafarad:
In commit 89cb8987, "feat: Changes transcript msgs to have timestamp as milliseconds," there are two issues:

1. This didn't change the JSON output for all transcript messages. For instance, START and END event messages continued to have their timestamps encoded as an ISO-8601 string.
2. Arguably, most JSON parsers would expect epoch milliseconds encoded as a number rather than a string. 

This humble PR addresses both of those issues by ensuring all transcript JSON messages have their timestamps encoded as a JSON number (milliseconds since epoch).

### DIFF
--- a/src/main/java/org/jitsi/jigasi/JvbConference.java
+++ b/src/main/java/org/jitsi/jigasi/JvbConference.java
@@ -1655,7 +1655,8 @@ public class JvbConference
                 }
             }
             else if ("org.jitsi.jigasi.xmpp.acc.USER_ID".equals(overridenProp)
-                && JigasiBundleActivator.getConfigurationService().getBoolean(overridePrefix + ".UNIQUE_USER_ID", false))
+                && JigasiBundleActivator.getConfigurationService().getBoolean(overridePrefix + ".UNIQUE_USER_ID",
+                    false))
             {
                 try
                 {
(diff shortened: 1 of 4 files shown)


damencho:
Have you tested this with jitsi-meet?

megafarad:
Yes. I verified jigasi sends milliseconds since epoch as JSON numbers for all Jitsi Meet meeting events.

damencho:
> Yes. I verified jigasi sends milliseconds since epoch as JSON numbers for all Jitsi Meet meeting events.

I mean, whether you have tested it with jitsi-meet, does the subtitles continue to work without a problem?

megafarad:
Ah! Yes, I just verified that they continue to work without a problem. 

At first, I didn't know because I couldn't get subtitles to work. I just enabled transcription in the jitsi-meet config (transcription.enabled) and saw the subtitles.

damencho:
Thank you for confirming. I will take a look this week.

Have you signed the CLA? (https://jitsi.org/icla for individuals and https://jitsi.org/ccla for corporations)

megafarad:
Just signed it.

damencho:
Can you please fix the warning about long lines, thank you?

damencho:
Hum, you do not have changes in that file. Can you rebase to latest maybe?

megafarad:
I'm definitely based on the latest.

As for the file: JvbConference.java, it looks like the long line was introduced by commit 3de47d0. The job for that commit's PR failed as well, with the same error (https://github.com/jitsi/jigasi/actions/runs/6581239749/job/18013060515?pr=506).

damencho (approved):

PR #518, opened 2024-02-13
fix: transcriber settings precedence

rpurdel:


src/main/java/org/jitsi/jigasi/TranscriptionGateway.java
```diff
@@ -106,15 +106,17 @@ private String getCustomTranscriptionServiceClass(String tenant)
                         REMOTE_TRANSCRIPTION_CONFIG_URL,
                         null);
 
-        if (remoteTranscriptionConfigUrl != null)
+        if (remoteTranscriptionConfigUrl != null && tenant != null)
         {
+            logger.info("Retrieving transcriber from remote URL " + remoteTranscriptionConfigUrl + "/" + tenant);
```
damencho, line 111: Do we need that log on every conference?

rpurdel: moved to debug

src/main/java/org/jitsi/jigasi/TranscriptionGateway.java
```diff
@@ -133,24 +140,36 @@ private String getCustomTranscriptionServiceClass(String tenant)
...
-        return JigasiBundleActivator.getConfigurationService()
+        logger.info("Attempting to retrieve transcriber from local configuration.");
+        String customTranscriptionServiceClass
+                = JigasiBundleActivator.getConfigurationService()
                 .getString(
                         CUSTOM_TRANSCRIPTION_SERVICE_PROP,
                         null);
+        if (customTranscriptionServiceClass != null)
+        {
+            logger.info("Using " + customTranscriptionServiceClass + " as the transcriber class.");
```
damencho, line 166: Maybe leave only the print if there is a tenant enabling non default transcription. 
And please add the callContext to the log as the rest of the code.

src/main/java/org/jitsi/jigasi/TranscriptionGateway.java
```diff
@@ -133,24 +140,36 @@ private String getCustomTranscriptionServiceClass(String tenant)
...
+        }
+        else
+        {
+            logger.info("No custom transcriber class found, falling back to default.");
```
damencho, line 170: We do not need this log.

damencho (approved):

PR #519, opened 2024-02-13
Transcribing backend flag

damencho:


src/main/java/org/jitsi/jigasi/JvbConference.java
```diff
@@ -729,6 +724,50 @@ else if (evt.getNewState() == RegistrationState.CONNECTION_FAILED)
...
+                        .discoverInfo(JidCreate.domainBareFrom(this.callContext.getRoomJidDomain()));
+
+                DiscoverInfo.Identity avIdentity = info.getIdentities().stream().
+                    filter(di -> di.getCategory().equals("component") && di.getType().equals("av_moderation"))
+                        .findFirst().orElse(null);
+
+                if (avIdentity != null && this.getAudioModeration() != null)
+                {
+                    String avModerationAddress = avIdentity.getName();
+                    this.getAudioModeration().setAvModerationAddress(avModerationAddress);
```
bgrozev, line 764: I think you need a null check, `this.audioModeration` is set to null in some cases

src/main/java/org/jitsi/jigasi/JvbConference.java
```diff
@@ -729,6 +724,50 @@ else if (evt.getNewState() == RegistrationState.CONNECTION_FAILED)
...
+
+                    logger.info(String.format("%s Discovered %s for %oms.",
```
bgrozev, line 766: Do we really care about this message only for `avModarationAddress`? I suggest either remove it or move it out of the "if" to be printed right after the `discoverInfo` call

src/main/java/org/jitsi/jigasi/JvbConference.java
```diff
@@ -760,6 +766,24 @@ private void discoverComponentAddresses()
...
                 }
+
+                DiscoverInfo.Identity roomMetadataIdentity = info.getIdentities().stream().
+                        filter(di -> di.getCategory().equals("component") && di.getType().equals("room_metadata"))
+                        .findFirst().orElse(null);
+
+                // we process room metadata messages only when we are transcribing
+                if (roomMetadataIdentity != null && this.gatewaySession instanceof TranscriptionGatewaySession)
+                {
+                    if (roomMetadataListener != null)
```
bgrozev, line 777: Any reason not to make the field `final` and always initialize it? You can remove these 5 lines of code and not worry about this executing after a re-connect for example

src/main/java/org/jitsi/jigasi/TranscriptionGatewaySession.java
```diff
@@ -308,8 +320,14 @@ void notifyChatRoomMemberUpdated(ChatRoomMember chatMember, Presence presence)
...
 
-        if (transcriber.isTranscribing() &&
-            !transcriber.isAnyParticipantRequestingTranscription())
+        this.maybeStopTranscription();
+    }
+
+    private void maybeStopTranscription()
+    {
+        if (transcriber.isTranscribing()
+                && !(transcriber.isAnyParticipantRequestingTranscription()
```
bgrozev, line 329: nit: suggest to extract a `isTranscriptionRequested()` method

src/main/java/org/jitsi/jigasi/JvbConference.java
damencho: But there is a check in the if?

src/main/java/org/jitsi/jigasi/JvbConference.java
damencho: I'll remove that and move the print after, I just kept it when moving the stuff.

src/main/java/org/jitsi/jigasi/JvbConference.java
bgrozev: Ops, I missed that

bgrozev (approved):

PR #520, opened 2024-02-16
fix: wrong transcription order

ltorje-8x8:
The receiving timestamp will be sent to skynet to be used as absolute timestamp for the audio chunk - the network lag is presumed to be less than tens of milliseconds (a latency of 20-50 ms is expected and it is considered to be ok as loss of time).

### DIFF
--- a/src/main/java/org/jitsi/jigasi/transcription/WhisperWebsocket.java
+++ b/src/main/java/org/jitsi/jigasi/transcription/WhisperWebsocket.java
@@ -17,33 +17,23 @@
  */
 package org.jitsi.jigasi.transcription;
 
-import io.jsonwebtoken.JwtBuilder;
-import io.jsonwebtoken.Jwts;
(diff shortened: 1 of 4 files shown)


damencho (commented):
Please fix the imports to be star imports.
Please avoid making style changes in the same commit with some fixes as it make it impossible for review, so can you isolate the fix in separate commit.

About the style of the java code all existing code follow
https://desktop.jitsi.org/Documentation/CodeConvention.html
And  we tried to enforce it with the checkstyle.

src/main/java/org/jitsi/jigasi/transcription/WhisperWebsocket.java
```diff
@@ -33,25 +33,26 @@
 import java.time.*;
 import java.util.*;
 import java.util.concurrent.*;
+import java.util.function.Supplier;
```
damencho, line 36: Star import?

PR #526, opened 2024-03-11
latest_long as the default google ts model

rpurdel:


saghul:
Can't we make this a config option, defaulting to that new model? It feels like it would make experimenting easier.

rpurdel:
> Can't we make this a config option, defaulting to that new model? It feels like it would make experimenting easier.

Done

saghul (dismissed):

bgrozev (dismissed):

damencho (approved):

PR #532, opened 2024-04-16
feat: transcriber detect bots and leave

damencho:


src/test/java/org/jitsi/jigasi/CallsHandlingTest.java
```diff
@@ -557,6 +587,8 @@ public void testNoFocusInTheRoom()
 
         AbstractGateway.setJvbInviteTimeout(
             AbstractGateway.DEFAULT_JVB_INVITE_TIMEOUT);
+
+        logger.info("");
```
bgrozev, line 591: Add a message?

src/main/java/org/jitsi/jigasi/util/Util.java
```diff
@@ -226,4 +262,80 @@ public static ExtensionElement createFeature(String var)
...
+                = JigasiBundleActivator.getConfigurationService().getString(P_NAME_TRUSTED_DOMAINS);
+
+            if (trustedDomainsStr != null)
+            {
+                JSONParser parser = new JSONParser();
+                try
+                {
+                    JSONArray trustedArray = (JSONArray) parser.parse(trustedDomainsStr);
+
+                    Util.trustedDomains = new ArrayList<String>(trustedArray);
```
bgrozev, line 281: Suggest logging the trusted domain setting when it's initialized

bgrozev (approved):

PR #542, opened 2024-08-20
fix: set the timestamps from jigasi

rpurdel:


oanaianc (dismissed):

damencho:
This will not give correct results as sometimes the results can be not in the correct order - partial and final results.
We had similar problem in google transcriptions where we were creating the timestamp ids on receiving the results. We needed to move it to when the audio is being received, before sent to the transcriber service, this way the timestamped ids are guranteed to be in correct order as this is the time when audio is received in jigasi.

rpurdel:
> This will not give correct results as sometimes the results can be not in the correct order - partial and final results. We had similar problem in google transcriptions where we were creating the timestamp ids on receiving the results. We needed to move it to when the audio is being received, before sent to the transcriber service, this way the timestamped ids are guranteed to be in correct order as this is the time when audio is received in jigasi.

The Whisper backend always returns the transcriptions in the correct order.

damencho (approved):

damencho:
Can you rebase as ot has conflicts

damencho (approved):

PR #556, opened 2024-09-30
Refactor transcriber selection

rpurdel:

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,7 +146,7 @@ Jigasi supports multiple transcription services, including Google Cloud speech-t
 API, Vosk speech recognition server, a custom flavor of Whisper
 and Oracle Cloud AI Speech.
 
-Google configuration
+Google configuration for transcription
 ====================
 
 For Jigasi to act as a transcriber, it sends the audio of all participants in the
@@ -180,11 +180,6 @@ In the file `/etc/jitsi/jigasi/config` add at the end:
 GOOGLE_APPLICATION_CREDENTIALS=/etc/jitsi/jigasi/application_default_credentials.json
 ```
 
-Set the following property in `~/jigasi/jigasi-home/sip-communicator.properties`:
-
-```
-org.jitsi.jigasi.transcription.customService=GOOGLE
-```
 and restart Jigasi.
 
 Vosk configuration for transcription
@@ -197,13 +192,13 @@ start the server with a docker:
 docker run -d -p 2700:2700 alphacep/kaldi-en:latest
 ```
 
-Then configure the transcription class with the following property in `~/jigasi/jigasi-home/sip-communicator.properties`:
+Then configure the transcription class with the following property in `/etc/jitsi/jigasi/sip-communicator.properties`:
 
 ```
-org.jitsi.jigasi.transcription.customService=VOSK
+org.jitsi.jigasi.transcription.customService=org.jitsi.jigasi.transcription.VoskTranscriptionService
 ```
 
-Finally, configure the websocket URL of the VOSK service in `~/jigasi/jigasi-home/sip-communicator.properties`:
+Finally, configure the websocket URL of the VOSK service in `/etc/jitsi/jigasi/sip-communicator.properties`:
 
 If you only have one instance of VOSK server:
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -152,19 +163,29 @@ gcloud init
...
+
 In the file `/etc/jitsi/jigasi/config` add at the end:
+
 ```
 # Credential for Google Cloud Speech API
 GOOGLE_APPLICATION_CREDENTIALS=/etc/jitsi/jigasi/application_default_credentials.json
 ```
-Restart jigasi
+
+Set the following property in `~/jigasi/jigasi-home/sip-communicator.properties`:
```
damencho, line 183: Why this file in the docs in rest of the places we refer to `/etc/jitsi/jigasi/sip-communicator.properties`?

README.md
```diff
@@ -152,19 +163,29 @@ gcloud init
...
+
+```
+org.jitsi.jigasi.transcription.customService=GOOGLE
```
damencho, line 186: Why do we need this at all? Right now the default is Google and we have not changed that.

README.md
```diff
@@ -176,18 +197,18 @@ start the server with a docker:
 docker run -d -p 2700:2700 alphacep/kaldi-en:latest
 ```
 
-Then configure the transcription class with the following properly in `~/jigasi/jigasi-home/sip-communicator.properties`:
+Then configure the transcription class with the following property in `~/jigasi/jigasi-home/sip-communicator.properties`:
```
damencho, line 200: Yeah probably we need to change and this one to reference the config in /etc/jitsi/jigasi.

README.md
```diff
@@ -176,18 +197,18 @@ start the server with a docker:
...
 
 ```
-org.jitsi.jigasi.transcription.customService=org.jitsi.jigasi.transcription.VoskTranscriptionService
```
damencho, line 182: Will the old one continue to work? As it will break people deployments once they update.

README.md
```diff
@@ -196,6 +217,36 @@ the URLs of different VOSK instances in JSON format:
 # org.jitsi.jigasi.transcription.vosk.websocket_url={"en": "ws://localhost:2700", "fr": "ws://localhost:2710"}
 ```
 
+Whisper configuration for transcription
+==================
+
+If you plan to use our own flavor of Whisper (check [jitsi/skynet](github.com/jitsi/skynet)), start by configuring the 
+following properties in `~/jigasi/jigasi-home/sip-communicator.properties`:
```
damencho, line 224: Same, etc folder.

README.md
```diff
@@ -196,6 +217,36 @@ the URLs of different VOSK instances in JSON format:
...
+If you also plan to enable the ASAP authentication, have a look at the 
+[documentation](https://github.com/jitsi/skynet/blob/master/docs/streaming_whisper_module.md) and at the properties 
+in the transcription options section of this README.
+
+
+Oracle Cloud AI Speech configuration for transcription
+==================
+
+To use [Oracle Cloud AI Speech](https://docs.oracle.com/en-us/iaas/Content/speech/home.htm), you need to configure the 
+following properties in `~/jigasi/jigasi-home/sip-communicator.properties`:
```
damencho, line 240: Same.

README.md
```diff
@@ -196,6 +217,36 @@ the URLs of different VOSK instances in JSON format:
...
+
+```
+org.jitsi.jigasi.transcription.customService=ORACLE_CLOUD_AI_SPEECH
+org.jitsi.jigasi.transcription.oci.websocketUrl=wss://realtime.aiservice-<<ENV>>.<<REGION>>.oci.oraclecloud.com
+```
+
+You also need to place valid OCI credentials under `~/.oci`.
```
damencho, line 247: Hum, by default jigasi runs as user jigasi and group jitsi and there is no home folder created for it, so that will not work with the debian packages. Isn't there another way to configure that?

src/main/java/org/jitsi/jigasi/TranscriptionGateway.java
```diff
@@ -92,35 +140,51 @@ public void stop()
...
-                        null);
 
-        if (remoteTranscriptionConfigUrl != null && tenant != null)
+        if (remoteTranscriptionConfigUrl != null)
         {
-            String tsConfigUrl = remoteTranscriptionConfigUrl + "/" + tenant;
+            String tsConfigUrl;
+
+            // this is JaaS specific
+            if (remoteTranscriptionConfigUrl.contains("jitsi.net"))
```
damencho, line 162: Why do we need that, can't we use the same format of the URL always?

src/main/java/org/jitsi/jigasi/TranscriptionGateway.java
```diff
@@ -92,35 +140,51 @@ public void stop()
...
         if (transcriberClass == null)
         {
-            transcriberClass
-                    = JigasiBundleActivator.getConfigurationService()
+            String transcriberType = JigasiBundleActivator.getConfigurationService()
                     .getString(
                             CUSTOM_TRANSCRIPTION_SERVICE_PROP,
                             null);
+            transcriberClass = transcriberClasses.getOrDefault(transcriberType, null);
+            logger.info("Transcriber class retrieved from config: " + transcriberClass);
```
damencho, line 185: So this will break existing configuration that use customService configured with a class, we need to make sure we support both or create a new one and deprecate the old one printing warnings and give time for people to adapt, for several months or an year, as we release not so often and that property was there for 5-6 years maybe.

damencho (changes requested):

src/main/java/org/jitsi/jigasi/TranscriptionGateway.java
rpurdel: Not really. As you some of the clients will have the ability to choose a specific transcriber on a per room basis.

damencho: Can we always do /tenanat?roomname= ... Hardcoding dns here is bot something we want.

PR #557, opened 2024-10-02
Flush transcription buffer on mute

rpurdel:

### DIFF
--- a/src/main/java/org/jitsi/jigasi/transcription/Participant.java
+++ b/src/main/java/org/jitsi/jigasi/transcription/Participant.java
@@ -101,16 +101,6 @@ public class Participant
     private Transcriber transcriber;
 
     /**
-     * The time when the last audio packet was received
-     */
-    private long lastAudioReceivedTime = System.currentTimeMillis();
-
-    /**
-     * Whether the buffer has been flushed after a period of inactivity
-     */
-    private boolean flushed = false;
-
-    /**
      * The chat room participant.
      */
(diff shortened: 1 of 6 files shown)


src/main/java/org/jitsi/jigasi/transcription/Participant.java
```diff
@@ -766,4 +779,36 @@ public boolean isRequestingTranscription()
 
         return ext != null && Boolean.parseBoolean(ext.getText());
     }
+
+    private void startBufferCheck()
+    {
+        new Thread(() -> {
```
damencho, line 785: This thread is not a daemon thread and will be preventing JVM from exiting. The code formatting is not correct. 
And in general, adding a thread for every participant even if the participant will not be transcribed will increase the threads drastically and is a waste of resources. 
Let's discuss this and maybe we can find a better approach.

damencho: Maybe we can add one global thread that goes over transcriber sessions and go over participants every second, using the scheduled executors and flush when needed.
It is not time consuming task and there will very few to flush.

damencho: I wonder though, that the buffer is not full which means that is less than 500ms of speech ... how many words is this ... I don't think that is more than one word. But yeah it may be the end of one and another one, so it will prevent a total of two words from being transcribed.

PR #564, opened 2024-10-28
feat: Adds notification for conference audio recordings.

damencho:
Depends on https://github.com/jitsi/jitsi-xmpp-extensions/pull/109

### DIFF
--- a/src/main/java/org/jitsi/jigasi/sounds/SoundNotificationManager.java
+++ b/src/main/java/org/jitsi/jigasi/sounds/SoundNotificationManager.java
@@ -235,28 +235,35 @@ public class SoundNotificationManager
     {
         if (gatewaySession.getFocusResourceAddr().equals(presence.getFrom().getResourceOrEmpty().toString()))
         {
+            boolean isJibriRecordingOn = false;
             RecordingStatus rs = presence.getExtension(RecordingStatus.class);
-
             if (rs != null)
             {
-                notifyRecordingStatusChanged(rs.getRecordingMode(), rs.getStatus());
+                isJibriRecordingOn = true;
+            }
 
+            boolean isAudioRecordingOn = false;
+            ConferenceProperties props = presence.getExtension(ConferenceProperties.class);
+
+            if (props != null)
+            {
+                ConferenceProperties.ConferenceProperty prop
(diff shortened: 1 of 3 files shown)


src/main/java/org/jitsi/jigasi/sounds/SoundNotificationManager.java
```diff
@@ -232,13 +233,30 @@ private CallContext getCallContext()
...
-            && gatewaySession.getFocusResourceAddr().equals(
-                presence.getFrom().getResourceOrEmpty().toString()))
+        if (gatewaySession.getFocusResourceAddr().equals(presence.getFrom().getResourceOrEmpty().toString()))
         {
-            notifyRecordingStatusChanged(rs.getRecordingMode(), rs.getStatus());
+            RecordingStatus rs = presence.getExtension(RecordingStatus.class);
+
+            if (rs != null)
+            {
+                notifyRecordingStatusChanged(rs.getRecordingMode(), rs.getStatus());
```
bgrozev, line 242: I think this will result in playing "recording is off" when both jibri and audio recording are on and the jibri is stopped. I think we should read the recording state from both the jibri extension and ConferenceProperties together, and then act on it.

src/main/java/org/jitsi/jigasi/sounds/SoundNotificationManager.java
```diff
@@ -232,13 +233,37 @@ private CallContext getCallContext()
...
+            ConferenceProperties props = presence.getExtension(ConferenceProperties.class);
+
+            if (props != null)
+            {
+                ConferenceProperties.ConferenceProperty prop
+                    = props.getProperties().stream()
+                        .filter(p -> ConferenceProperties.KEY_AUDIO_RECORDING_ENABLED.equals(p.getKey()))
+                        .findFirst().orElse(null);
+
+                isAudioRecordingOn = prop != null && prop.getValue().equals(Boolean.TRUE.toString());
```
bgrozev, line 255:
```suggestion
                isAudioRecordingOn = prop != null && Boolean.parseBoolean(prop.getValue());
```

bgrozev (approved):

PR #566, opened 2024-10-30
fix(whisper): Reconnects on socket closed while operating.

damencho:
Executes stop and connect in a different from the websocket messages thread as recommended in jetty docs.

### DIFF
--- a/src/main/java/org/jitsi/jigasi/transcription/WhisperConnectionPool.java
+++ b/src/main/java/org/jitsi/jigasi/transcription/WhisperConnectionPool.java
@@ -61,7 +61,6 @@ public class WhisperConnectionPool
             logger.info("Room " + roomId + " doesn't exist. Creating a new connection.");
             final WhisperWebsocket socket = new WhisperWebsocket();
 
-            // connect socket in new thread to not block Smack threads
             socket.connect();
 
             pool.put(roomId, socket);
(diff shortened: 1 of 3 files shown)


src/main/java/org/jitsi/jigasi/transcription/WhisperConnectionPool.java
```diff
@@ -67,7 +62,7 @@ public WhisperWebsocket getConnection(String roomId)
             final WhisperWebsocket socket = new WhisperWebsocket();
 
             // connect socket in new thread to not block Smack threads
```
bgrozev, line 64: Are we not worried about blocking Smack threads anymore? If not, update the doc :)

src/main/java/org/jitsi/jigasi/transcription/WhisperConnectionPool.java
```diff
@@ -83,29 +78,19 @@ public WhisperWebsocket getConnection(String roomId)
     public void end(String roomId, String participantId)
     {
         // execute this in new thread to not block Smack
```
bgrozev, line 80: Same

damencho: It is done inside those methods.

src/main/java/org/jitsi/jigasi/transcription/WhisperConnectionPool.java
damencho: It is done inside those methods.

bgrozev (approved):

PR #568, opened 2024-10-31
Process translations and xmpp sending of transcription in separate threads

damencho:

### DIFF
--- a/src/main/java/org/jitsi/jigasi/transcription/TranslationManager.java
+++ b/src/main/java/org/jitsi/jigasi/transcription/TranslationManager.java
@@ -17,7 +17,10 @@
  */
 package org.jitsi.jigasi.transcription;
 
+import org.jitsi.jigasi.util.Util;
+
 import java.util.*;
+import java.util.concurrent.*;
 
 /**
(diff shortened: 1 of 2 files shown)


src/main/java/org/jitsi/jigasi/transcription/AbstractTranscriptPublisher.java
```diff
@@ -157,14 +159,29 @@ public abstract class AbstractTranscriptPublisher<T>
         = Logger.getLogger(AbstractTranscriptPublisher.class);
 
     /**
-     * Aspect for successful upload of transcript
+     * A queue used to offload xmpp message sending in a new thread to avoid blocking.
      */
-    private static final String DD_ASPECT_SUCCESS = "upload_success";
+    private static final PacketQueue<Runnable> xmppInvokeQueue = new PacketQueue<>(
```
bgrozev, line 164: This being static means that order will be preserved for everything. I it makes more sense to have a separate queue for each AbstractTranscriptPublisher

bgrozev (approved):

PR #575, opened 2024-12-11
ref: Adapt to jicoco changes, move CmdLine to jigasi.

bgrozev:


damencho (approved):

damencho:
I wonder do we still need those cmd args thingy ...

damencho:
And there is a problem with dependencies

damencho (approved):

PR #581, opened 2025-01-26
Update WhisperWebsocket.java

behzadCube:
check privateKey and privateKeyName contain.
if the variable imported from environment variable must use ;


damencho:
Can you share what is the issue you fix? How to try reproduce it?

behzadCube:
I trying to connect jigasi for live transcription with skynet (with Whisper socket connection ). I deployed this services with docker compose and create share network between services but get Error on creating jwt token in situation that set AuthenticationByPass= true 
this is logs 
image

prasadvempower:
@damencho - Please accept this pull request. I am also facing same problem

damencho:

damencho (changes requested):
@behzadCube Can you fix the formatting, please?

felipeavilis:
Facing the same problem here.

PR #586, opened 2025-03-21
fix: clean up XMPP resource id

hamletmun:
Change replace to replaceAll
See https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/String.html#replaceAll(java.lang.String,java.lang.String)

CLA (https://jitsi.org/icla for individuals) already signed


hamletmun:
https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/String.html
``` java
public String replaceAll​(String regex, String replacement)
public String replace​(CharSequence target, CharSequence replacement)
```

damencho (approved):

damencho (approved):

PR #594, opened 2025-05-08
feat: Uses metadata to get startmuted value.

damencho:

### DIFF
--- a/src/main/java/org/jitsi/jigasi/JvbConference.java
+++ b/src/main/java/org/jitsi/jigasi/JvbConference.java
@@ -1141,7 +1141,7 @@ public class JvbConference
             // let's check room config
             updateFromRoomConfiguration();
 
-            logger.info("Joined room: " + roomName + " meetingId:" + this.getMeetingId());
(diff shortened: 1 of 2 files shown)


src/main/java/org/jitsi/jigasi/JvbConference.java
```diff
@@ -2250,10 +2251,7 @@ private void updateFromRoomConfiguration()
 
     private void processRoomMetadataJson(String json)
```
bgrozev, line 2252: The catch block used to span the entire method, now it only covers the json parser. Is this intended and safe?

damencho: yep. I have added one default value that was missing. So it should be safe now.

bgrozev (approved):

src/main/java/org/jitsi/jigasi/JvbConference.java
bgrozev: there's a bunch of other possible exceptions here, for example "startMuted: 0" will cause the cast to fail

PR #613, opened 2025-11-10
feat(transcriber): Moves to use speech to text v2.

damencho:


sliwken:
@damencho, thank you very much fro this change, I saw a new configuration GOOGLE_LOCATION to set location in recognizer resource name.
Regional API endpoint URL must match regional location (e.g. https://us-central1-speech.googleapis.com for the us-central1 location).
Could you also put a configuration to set regional endpoint when creating `SpeechClient` instance
```
SpeechSettings speechSettings = SpeechSettings.newBuilder().setEndpoint("us-central1-speech.googleapis.com:443").build();
SpeechClient client = SpeechClient.create(speechSettings);
```

PR #619, opened 2025-12-09
fix thread leaks in WhisperWebsocket

rgumirov00:
fix thread leaks in WhisperWebsocket

if WhisperWebsocket return error Jigasi experiences a thread leak. Each call creates about 120 threads that never get cleaned up after a Whisper failure

After 100 calls my jigasi use only 80 threads!
Before fix it was about 8000-10000 threads


rgumirov00:
> Hi, thanks for your contribution! If you haven't already done so, could you please make sure you sign our CLA (https://jitsi.org/icla for individuals and https://jitsi.org/ccla for corporations)? We would unfortunately be unable to merge your patch unless we have that piece :(.

done

Issue #635, opened 2026-04-20
Translation only works one-way: PL→others works, but EN/DE→PL does not work


gmus-dev:
## Environment
- Jitsi Meet: 2.0.10888-1
- Jigasi: 1.1.407-g1e3a7ac
- Jicofo: 1.0-1174-1
- Prosody: 13.0.4
- Transcription: Google Cloud Speech-to-Text (latest_long model)
- Translation: Google Cloud Translation API

## Expected behavior
In a multilingual meeting:
- Polish speaker talks → other participants see subtitles translated to THEIR language ✅
- English speaker talks → Polish participant sees subtitles translated to Polish ❌
- German speaker talks → Polish participant sees subtitles translated to Polish ❌

Translation should work in ALL directions for ALL participants.

## Actual behavior
Translation only works ONE WAY:
- Polish speaker → English participant sees English subtitles ✅
- English speaker → Polish participant sees RAW ENGLISH TEXT (no translation) ❌

When Polish participant selects "Polish" as their subtitle language and the 
English speaker talks, Jigasi throws:

UnsupportedOperationException: pl is not a language supported 
by the Google Cloud speech-to-text API

This means Jigasi is incorrectly passing `translation_language=pl` 
(the TARGET language for translation) into the STT language validator
`validateLanguageTag()` instead of passing it only to TranslationManager.

## Root cause (suspected)
In GoogleCloudTranscriptionService.java, the `validateLanguageTag()` 
function is called with the value of `translation_language` presence 
property (e.g. "pl", "en") which are short 2-letter codes.

These short codes are NOT in SUPPORTED_LANGUAGE_TAGS (which requires 
full BCP-47 codes like "pl-PL", "en-US"), so the exception is thrown.

`translation_language` should only be passed to TranslationManager 
as the TARGET language for translation — it should NEVER be used 
as a STT language code.

## Config
config.js:
transcription: {
    enabled: true,
    translationLanguages: ['pl', 'en', 'de', 'uk'],
    translationLanguagesHead: ['pl'],
    useAppLanguage: false,
    preferredLanguage: 'pl-PL',
},

sip-communicator.properties:
org.jitsi.jigasi.ENABLE_TRANSCRIPTION=true
org.jitsi.jigasi.transcription.customService=...GoogleCloudTranscriptionService
org.jitsi.jigasi.transcription.ENABLE_TRANSLATION=true
org.jitsi.jigasi.transcription.translationService=...GoogleCloudTranslationService

[...]

PR #639, opened 2026-06-01
Resolve bare locale codes to supported BCP-47 language tags in recognition config

saurabhffs:
**Root Cause Analysis (RCA) Issue #635 **

Jigasi’s expected flow is:

 - transcribe incoming speech in the speaker’s language,
 - translate the finalized transcript into each requested target language,
 - publish the translated result to the participant/client that requested that language.

The bug occurred because the reverse path for Polish as a target language was not equivalent to the path where Polish was the source language. In effect, the system handled PL -> other correctly, but other -> PL was not going through the same valid target-language resolution/request path, which resulted in no Polish translation output even though transcription itself was available.

**Fix Implementation**
 - Added strict language code normalization in the translation request builder to ensure frontend locale strings (e.g., pl-PL, pl_PL) are correctly mapped to the 2-letter ISO code (pl) required by the translation backend.

 - Improved logging in the TranslationManager to surface HTTP 400s or unrecognized target language exceptions from the translation provider instead of failing silently.

### DIFF
--- a/src/main/java/org/jitsi/jigasi/transcription/GoogleCloudTranscriptionService.java
+++ b/src/main/java/org/jitsi/jigasi/transcription/GoogleCloudTranscriptionService.java
@@ -265,6 +265,26 @@ public class GoogleCloudTranscriptionService
 
         // set the Language tag
         String languageTag = request.getLocale().toLanguageTag();
+        logger.debug("PATCH ACTIVE: languageTag=" + languageTag);
+        if (!languageTag.contains("-"))
+        {
+            // try java's own Locale resolution first
+            String resolvedTag = new Locale(languageTag).toLanguageTag();
+
+            // if still bare (e.g., "pl"), look up in SUPPORTED_LANGUAGE_TAGS
+            if (!resolvedTag.contains("-"))
+            {
+                String prefix = resolvedTag.toLowerCase();
+                resolvedTag = Arrays.stream(SUPPORTED_LANGUAGE_TAGS)
+                        .filter(t -> t.toLowerCase().startsWith(prefix + "-"))
+                        .findFirst()
+                        .orElseThrow(() -> new UnsupportedOperationException(
+                                prefix + " could not be resolved to a supported BCP-47 tag"));
+            }
+
+            languageTag = resolvedTag;
+        }
+
         validateLanguageTag(languageTag);
         builder.setLanguageCode(languageTag);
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/jitsi/jigasi/transcription/GoogleCloudTranscriptionService.java
```diff
@@ -265,6 +265,26 @@ private RecognitionConfig getRecognitionConfig(TranscriptionRequest request)
 
         // set the Language tag
         String languageTag = request.getLocale().toLanguageTag();
+        logger.debug("PATCH ACTIVE: languageTag=" + languageTag);
+        if (!languageTag.contains("-"))
+        {
+            // try java's own Locale resolution first
+            String resolvedTag = new Locale(languageTag).toLanguageTag();
```
damencho: Isn't this done on line 267? 
`String languageTag = request.getLocale().toLanguageTag();`

saurabhffs: this seems to be a redundant line of code. I will incorporate the change.

damencho:

saurabhffs:
signed

src/main/java/org/jitsi/jigasi/transcription/GoogleCloudTranscriptionService.java
```diff
@@ -265,6 +265,18 @@ private RecognitionConfig getRecognitionConfig(TranscriptionRequest request)
 
         // set the Language tag
         String languageTag = request.getLocale().toLanguageTag();
+        logger.debug("PATCH ACTIVE: languageTag=" + languageTag);
```
damencho: Can you drop this?

damencho (changes requested):

damencho:
> signed

Seems you are using different email for your commits than the one in the CLA.

src/main/java/org/jitsi/jigasi/transcription/GoogleCloudTranscriptionService.java
saurabhffs: sure.

saurabhffs:
No, this is the mail ending with ond@gmail.com

damencho:
> No, this is the mail ending with [ond@gmail.com](mailto:ond@gmail.com)

Nope: https://github.com/saurabhffs/jigasi/commit/b2e927c6e2c9c10b71093da4d6bf50093923c3b5.patch

saurabhffs:
this is an old college mail that is still associated with my github account. the email doesn't exist anymore. should I raise another PR after email change?

damencho:
> this is an old college mail that is still associated with my github account. the email doesn't exist anymore. should I raise another PR after email change?

yep, that will work

PR #647, opened 2026-09-10
fix(transcription): don't share a JSONParser across Vosk sessions

zoubenjia:
## Problem

`org.json.simple.parser.JSONParser` is **not thread-safe** — it keeps the lexer, the parse status and the handler status stack in instance fields and mutates them throughout `parse()`.

`VoskTranscriptionService` holds a single parser in a field of the **outer** class:

https://github.com/jitsi/jigasi/blob/master/src/main/java/org/jitsi/jigasi/transcription/VoskTranscriptionService.java#L79

while `VoskWebsocketStreamingSession` — **one instance per participant** — calls it from `@OnWebSocketMessage`:

https://github.com/jitsi/jigasi/blob/master/src/main/java/org/jitsi/jigasi/transcription/VoskTranscriptionService.java#L285

Jetty dispatches those callbacks concurrently on pool threads, so parses from different participants corrupt each other's state.

## Symptoms in production

Several speakers in one conference. Valid Vosk replies fail to parse, and the resulting `onError` tears that participant's session down, so they lose captions:

```
SEVERE: VoskWebsocketStreamingSession.onMessage: Error parsing message: {
  "partial" : "the out for the ancient"
}
Unexpected character (") at position 29.

SEVERE: VoskWebsocketStreamingSession.onError: Error while streaming audio data
        to transcription service
java.util.NoSuchElementException
  at VoskTranscriptionService.onMessageInternal(VoskTranscriptionService.java:286)
```

Note the logged JSON is perfectly valid — the parser state was already corrupted by another thread.

## It does not always throw

This is the part that worries me more. A stress test against the real class (8 threads × 20 000 Vosk-shaped `{"partial": "..."}` messages, parser obtained by reflection from an actual `VoskTranscriptionService` instance) gave:

| | result |
|---|---|
| before | `total=160000 ok=96597 bad=63403` |
| after | `total=160000 ok=160000 bad=0` |

Among the failures, besides `ParseException` / `NoSuchElementException` / `IndexOutOfBoundsException: Range [-21, -21 + 16405) out of bounds for length 16384`, there were results that were **structurally valid but carried another participant's text**. So a transcript can silently attribute speech to the wrong speaker, with nothing in the logs.

## Fix

Give each thread its own parser.

I reused a thread-confined parser rather than allocating one per message: `new JSONParser()` eagerly builds a `Yylex` holding a 16384-char (32 KB) buffer, and Vosk emits partial results several times a second *per participant*, so per-message allocation puts a lot of garbage on the hot path. `parse()` calls `reset()` on entry, so reuse on a confined thread is safe, and the Jetty pool is bounded so the retained set is bounded too.

Happy to switch to a plain local variable if you'd prefer the simpler form.

## Note

`generateWebsocketUrl()` in this same class already allocates its own local parser, which suggests the shared field was an oversight rather than a deliberate choice.

Found while running jigasi as a transcriber against a self-hosted Vosk server on `docker-jitsi-meet` `stable-9955` (jigasi `1.1-369-gb80d3df`); the affected lines are unchanged on `master`.


damencho:

zoubenjia:
Thanks for the quick response — the ICLA is signed.

Re the CI: the earlier `Java 17` failure was `CallsHandlingTest.testFocusLeftTheRoomWithNoResume` (`expected: <null> but was: <MockMUC@...>`), which is unrelated to this change — nothing under `src/test` references `VoskTranscriptionService`. It passed on re-run, so it looks flaky rather than caused by this patch.

`codecov/patch` is red because the added lines are the javadoc and the field declaration itself. Happy to add a unit test if you'd like one: it would spin up N threads against a single `VoskTranscriptionService` instance and assert every partial round-trips intact. That fails reliably on the current code (I measured 63403 bad out of 160000 results at 8 threads, including `NoSuchElementException`, `ParseException` on valid JSON, and — more worryingly — structurally valid results carrying another participant's text) and is clean with this change. Let me know if that's worth having in the suite.

Also happy to switch the `ThreadLocal` to a plain per-message local if you prefer the simpler form. The only reason I reused a thread-confined parser is that `new JSONParser()` eagerly allocates a `Yylex` holding a 16384-char buffer, and Vosk emits partial results several times a second per participant.

damencho:
> Thanks for the quick response — the ICLA is signed.

The email you used to sign and the one in the changes are different, can you address that.

zoubenjia:
Good catch — thanks. The commit was authored with my work address by mistake; the ICLA is signed under my personal one.

I've amended the commit to `Benjia Zou <zoubenjia@gmail.com>`, which matches the signature, and force-pushed. The tree hash is unchanged (`b29c2909…` before and after), so the code is byte-for-byte the same — only the author identity changed.

damencho (approved):

